PR #3, opened 2021-05-12
Cm/update readme

CheezItMan:
My updates to the README.  I updated routes to look like Flask instead of Rails, and added endpoints descriptions for put and delete requests. 

Also we should make the default branch be `master` because Learn only grades the master branch. 

@jbieniosek


jmaddox19 (approved):
LGTM! Thanks Chris!

PR #8, opened 2021-05-14
Update the way error requirements to be flexible

jmaddox19:
…and slightly less verbose

### DIFF
--- a/ada-project-docs/wave_01.md
+++ b/ada-project-docs/wave_01.md
@@ -25,22 +25,27 @@ It's crucial for all APIs to be able to handle errors. For every required endpoi
 
 If something goes wrong...
 - Your API should return an appropriate HTTP status code.
-- For POST and PUT requests, responses with 4XX response codes should also return an array of error messages to explain what is wrong with the request.
+- For POST and PUT requests, responses with 4XX response codes should also return a response body with some indication of what went wrong.
 
-For this project, the list of errors should be formatted like this:
+
+If something goes wrong...
+- Your API should return an appropriate HTTP status code.
+- For POST and PUT requests, responses with 4XX response codes should also return a response body with some indication of what went wrong.
+
+This could be something as simple as:
 
 ```json
-// for a video endpoint
+{
+        "details": "Invalid data"
+}
+```
+
+...or something slightly more complex like:
+```json
 [
    "title must be provided and it must be a string",
    "total_inventory must be provided and it must be a number"
 ]
-
-// for a rental endpoint
-[
-   "video_id must match a video in the video store database"
-]
-
 ```
 
 # Requirements
--- a/ada-project-docs/wave_01.md
+++ b/ada-project-docs/wave_01.md
@@ -37,7 +37,6 @@ For this project, the list of errors should be formatted like this:
 ]
 
 // for a rental endpoint
-
 [
    "video_id must match a video in the video store database"
 ]
--- a/ada-project-docs/wave_01.md
+++ b/ada-project-docs/wave_01.md
@@ -24,7 +24,7 @@ Our goal for this wave is to be able to do all CRUD actions for these two entiti
 It's crucial for all APIs to be able to handle errors. For every required endpoint described in this project, handle errors in this pattern.
 
 If something goes wrong, your API should return:
-- an appropriate [HTTP status code](http://billpatrianakos.me/blog/2013/10/13/list-of-rails-status-code-symbols/)
+- an appropriate [HTTP status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status)
 - For POST and PUT requests, responses with 4XX response codes should also return an array of error messages to explain what is wrong with the request.
 
 For this project, the list of errors should be formatted like this:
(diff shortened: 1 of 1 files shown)


ada-project-docs/wave_01.md
```diff
@@ -25,27 +25,22 @@ It's crucial for all APIs to be able to handle errors. For every required endpoi
...
-{
-  "errors": {
-    "available_inventory": [
-      "can't be blank",
-      "is not a number"
-    ]
-  }
-}
+// for a video endpoint
+[
```
anselrognlie: either insert a blank line here (between the comment and the code) or remove the blank line between the comment and the code at line 40. looks a little janky right now. (fwiw, i'd prefer removing the blank later)

```suggestion
// for a video endpoint

[
```

ada-project-docs/wave_01.md
```diff
@@ -25,27 +25,22 @@ It's crucial for all APIs to be able to handle errors. For every required endpoi
...
+   "title must be provided and it must be a string",
+   "total_inventory must be provided and it must be a number"
+]
 
-// ...or...
+// ...or...for a rental endpoint
```
anselrognlie: are these two styles the exhaustive list? if so can they be labelled similarly? 

```suggestion
// for a rental endpoint
```

ada-project-docs/wave_01.md
```diff
@@ -25,27 +25,22 @@ It's crucial for all APIs to be able to handle errors. For every required endpoi
...
```
anselrognlie: rather than the earlier change, I'd suggest removing this blank line so the comment hugs the following data.
```suggestion
```

ada-project-docs/wave_01.md
```diff
@@ -397,4 +392,4 @@ Status: `200`
 
 ## Tests
 
-There are no Pytest tests for this project.  There are Postman test scripts for all of these endpoints.  If you want to write your own tests for this project, we recommend using the tests in Task List as a template.
\ No newline at end of file
+There are no Pytest tests for this project.  There are Postman test scripts for all of these endpoints.  If you want to write your own tests for this project, we recommend using the tests in Task List as a template.
```
anselrognlie, line 399: I'm guessing this is an unintended change caused by your editor inserting a trailing newline? Don't know how easy it is to prevent that from happening. FWIW, text files _are_ supposed to end with newlines, but they get left off all the time... So I'm pro add newlines!

ada-project-docs/wave_01.md
```diff
@@ -25,27 +25,22 @@ It's crucial for all APIs to be able to handle errors. For every required endpoi
 
 If something goes wrong, your API should return:
 - an appropriate [HTTP status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status)
-- a list of errors
+- For POST and PUT requests, responses with 4XX response codes should also return an array of error messages to explain what is wrong with the request.
```
anselrognlie: The lack of parallel structure in the bullets really bothers me. I think a little tweaking could help, but it requires editing 26-28, which git won't let me. Suggested markdown in the following comment. It still makes it clear that there are two situations to keep in mind, and feels a bit more consistent.

anselrognlie: If something goes wrong...
- Your API should return an appropriate [HTTP status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status).
- For POST and PUT requests, responses with 4XX response codes should also return an array of error messages to explain what is wrong with the request.

anselrognlie (commented):
Haven't been following too closely, so I don't presume to be a good sign-off candidate here, but had some comments. Do with them as you wish!

ada-project-docs/wave_01.md
```diff
@@ -25,27 +25,22 @@ It's crucial for all APIs to be able to handle errors. For every required endpoi
...
+// ...or...for a rental endpoint
 
-{
-    "errors": [
-        "Not Found"
-    ]
-}
+[
+   "video_id must match a video in the video store database"
+]
```
audreyandoy: In Task List we had the students return a dictionary as the error message. Did we want to stay consistent with that? This could also not matter much at all as errors can be returned as either, if so, disregard this comment 😆  
```suggestion
{
        "details": "Invalid data"
}
```

audreyandoy (commented):
I like the verbage. I made a comment about the error message but could just be a non-issue.

ada-project-docs/wave_01.md
jmaddox19: not exhaustive, but with one more example, it could be.

ada-project-docs/wave_01.md
jmaddox19: That's a good point Audrey. I think I'm pretty ok with the requirements being different for this project. But to be fair I didn't take the time to write tests to verify error messages so maybe we should be ahead and make the requirements more simple?

jmaddox19: Just synced with @jbieniosek  and updated requirements here to be explicitly very flexible

jbieniosek (approved):
Looks good!

PR #10, opened 2021-05-17
Making 2 tests less strict

jbieniosek:


jmaddox19 (approved):
Thanks! Any reason not to change lines 372 and 860 as well?

PR #16, opened 2021-05-19
Scissors - Sneha Agarwal

sneha8412:


jmaddox19:
Great work!

It looks like this project is missing a .gitignore file so unfortunately the diff is so big I can't do inline comments.

I enjoy that you setup 3 files for the different sets of routes :)

I see in Learn that you noted that you completed endpoints for overdue and video history but I'm not seeing that code for some reason. I wonder if I'm missing it somehow or if maybe it didn't get pushed up.

sneha8412:
Hi Jared, thank you for your comments. 

I have the .gitignore on my end on VS code, does that mean that Git just doesn't upload the gitinore file coz hey all the secrets are in there? so in that case I need not do anything right? :)

I worked on the optional enhancements this is the update:
GET /rentals/overdue - complete - 
GET /videos/<id>/history -complete
I had to remove it the above 2 sections of code from my program temporarily because of some error I got in CLI , I remember having the code, but some complication happened and then I removed it, again I should have just # it out. I will try and find the code and put it back in. I can't recall if I committed it. :(
GET /customers/<id>/history - did not work on it
deployment - did not deploy

PR #21, opened 2021-05-20
Paper - Katrina Li

likatrina7:


app/models/customer.py
```diff
@@ -0,0 +1,22 @@
+from app import db
+from datetime import datetime
+
+class Customer(db.Model):
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    name = db.Column(db.String)
+    postal_code = db.Column(db.String)
```
CheezItMan, line 7: This is fine, I'm just noting that the postman tests expect it to be an integer.

app/customer_routes.py
```diff
@@ -0,0 +1,59 @@
...
+from app import db
+from app.models.customer import Customer
+from app.models.video import Video
+
+customers_bp = Blueprint("customers", __name__, url_prefix="/customers")
+
+@customers_bp.route("", methods=["GET"], strict_slashes=False)
+def get_all_customers():
+    customers = Customer.query.all()
+    response_body =[customer.customer_to_json() for customer in customers]
```
CheezItMan, line 12: Nice list comprehension

app/customer_routes.py
```diff
@@ -0,0 +1,59 @@
...
+    return jsonify(response_body), 200
+
+@customers_bp.route("", methods=["POST"], strict_slashes=False)
+def create_new_customer():
+    new_customer_data = request.get_json()
+    if new_customer_data.keys() >= {"name", "postal_code", "phone"}:
+        new_customer = Customer(**new_customer_data)
```
CheezItMan, line 19: I like the use of **

app/customer_routes.py
```diff
@@ -0,0 +1,59 @@
...
+        db.session.add(new_customer)
+        db.session.commit()
+        return jsonify(new_customer.customer_to_json()), 201
+    return jsonify({"details": "Invalid data"}), 400
+
+@customers_bp.route("/<int:customer_id>", methods=["GET"], strict_slashes=False)
+def get_single_customer(customer_id):
+    customer = Customer.query.get_or_404(customer_id)
```
CheezItMan, line 27: Nice use of `get_or_404`, my only criticism of it is that it doesn't provide much feedback to the user other than the 404 response code.

app/customer_routes.py
```diff
@@ -0,0 +1,59 @@
...
+    return jsonify(customer.customer_to_json()), 200
+
+@customers_bp.route("/<int:customer_id>", methods=["PUT"], strict_slashes=False)
+def update_single_customer(customer_id):
+    update_customer_data = request.get_json()
+    customer = Customer.query.get_or_404(customer_id)
+    if update_customer_data.keys() >= {"name", "postal_code", "phone"}:
```
CheezItMan, line 34: Very dry way to validate the request body.

app/rental_routes.py
```diff
@@ -0,0 +1,47 @@
+from app.models.video import Video
```
CheezItMan, line 1: I like how you broke up the routes file into separate files for each type of route.

app/rental_routes.py
```diff
@@ -0,0 +1,47 @@
...
+
+    customer = Customer.query.get_or_404(check_out_data["customer_id"])
+    video = Video.query.get_or_404(check_out_data["video_id"])
+    check_out_rental = Rental(**check_out_data)
+    if video.available_inventory > 0:
+        video.available_inventory -= 1
+        customer.videos_checked_out_count += 1
+        db.session.add(check_out_rental)
+        db.session.commit()
+        return make_response(check_out_rental.rental_to_json(), 200)
```
CheezItMan, line 29: I do suggest pushing the "business logic" of checking out a video into model methods.  This way the route file stays focused on the task of coordinating the request and response and the code dealing with the logic of checking out a video goes with the models.

CheezItMan (commented):
This is a really solid submission Katrina.  I had a few comments, but my only real suggestion is to move some of the "business logic" into the models.  That way the routes can just handle the request-response part and the actual dealing with videos, customers and rentals can be done in the models.  I think it's a better way for separation of concerns.  Let me know what questions you have.

Otherwise this is quite well done and you showed a real solid understanding of Flask and Python.

PR #24, opened 2021-05-21
Paper - Jen

jenniferlim07:


app/routes.py
```diff
@@ -1,2 +1,327 @@
...
+    rentals = Rental.query.all()
+    print("******", rentals)
+
+    # rentals_response = [rental.to_json() for rental in rentals]
+    rentals_response = []
+    for rental in rentals:
+        # print("***", rental)
+        # print("***", rental.id)
+        # print("***", rental.customer_id)
+        # print("***", rental.video_id)
```
CheezItMan:
```suggestion
```

app/models/customer.py
```diff
@@ -0,0 +1,27 @@
...
+
+class Customer(db.Model):
+    __tablename__ = "customer"
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    name = db.Column(db.String)
+    postal_code = db.Column(db.Integer)
+    phone = db.Column(db.String)
+    registered_at = db.Column(db.DateTime)
+    videos_checked_out_count = db.Column(db.Integer) #, default=1)
+    video = relationship("Rental", back_populates="customer")
```
CheezItMan, line 13: You're not really using this relationship, and it doesn't make sense for an attribute Video to be a rental.

This might work better and would give every Customer object a `.rentals` attribute, which is a list of the customer's rentals.

```suggestion
    rentals = db.relationship('Rental', backref='customer', lazy=True)
```

app/models/rental.py
```diff
@@ -0,0 +1,22 @@
...
+    __tablename__ = "rental"
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    customer_id = db.Column("customer", db.Integer, db.ForeignKey("customer.id"), primary_key=True)
+    video_id = db.Column("video", db.Integer, db.ForeignKey("video.id"), primary_key=True)
+    due_date = db.Column(db.DateTime)
+    status = db.Column(db.String)
+    video = relationship("Video", back_populates="customer")
+    customer = relationship("Customer", back_populates="video")
+
+    def to_json(self):
```
CheezItMan, line 16: This is a good model method, but I would suggest adding methods to do the business logic for checking out and checking in videos here as well.  You can add some class methods to do this.

app/models/video.py
```diff
@@ -0,0 +1,27 @@
...
+from sqlalchemy.orm import relationship
+
+class Video(db.Model):
+    __tablename__ = "video"
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    title = db.Column(db.String)
+    release_date = db.Column(db.String)
+    total_inventory = db.Column(db.Integer)
+    available_inventory = db.Column(db.Integer) #, default=1)
+    customer = relationship("Rental", back_populates="video")
```
CheezItMan, line 12: Again this relationship doesn't make sense as they're not directly connected.

app/routes.py
```diff
@@ -1,2 +1,327 @@
...
+customers_bp = Blueprint("customers", __name__, url_prefix="/customers")
+videos_bp = Blueprint("videos", __name__, url_prefix="/videos")
+rentals_bp = Blueprint("rentals", __name__, url_prefix="/rentals")
+
+
+def not_found():
+    return make_response("Not Found", 404)
+
+def invalid_data():
+    return jsonify({"details": "Invalid data"}), 400
```
CheezItMan, line 18: Good helper functions

app/routes.py
```diff
@@ -1,2 +1,327 @@
...
+    inner.__name__ = func.__name__
+    return inner
+
+def video_not_found(func):
+    def inner(id):
+        if Video.query.get(id) is None:
+            return make_response("Not Found", 404)
+        return func(id)
+    inner.__name__ = func.__name__
+    return inner
```
CheezItMan, line 35: These are great decorators.

app/routes.py
```diff
@@ -1,2 +1,327 @@
...
+    for rental in customer.video:
+        if rental.status == "checked-in":
+            return make_response({400: "Bad Request"}, 400)
+        if rental.video_id == video_id:
+            customer.videos_checked_out_count -= 1
+            video.available_inventory += 1
+            rental.status = "checked-in"
+            # db.session.delete(rental)
+
+    db.session.commit()
```
CheezItMan, line 275: I suggest putting the "check-in" business logic into the Rental model.  That way the `routes.py` file is only concerned with dealing with requests and responses and the Rental model can handle the logic of checking out and checking in videos.  It's good for separation of concerns and making the app more flexible.  Then you could theoretically reuse the models for a mobile app or any non-api application.

CheezItMan (commented):
Nice work Jen, you hit the main learning goals here.  Well done.   I really like what you did with the decorator functions, those really dried things up quite a bit.   I had a few suggestions on organization, and using relationships.  Let me know if you have questions on them.

PR #25, opened 2021-05-21
Scissors - Sumitra Chhetri 

sumitrac:


app/routes.py
```diff
@@ -1,2 +1,349 @@
...
+    
+    customer = Customer.query.get(customer_id)
+    video = Video.query.get(video_id)
+
+    for rental in customer.video:
+        if rental.video_id == video_id:
+            customer.checkout_count -= 1
+            video.available_inventory += 1
+            db.session.delete(rental)
+            db.session.commit()
```
jmaddox19, line 242: This is interesting that there's a loop here. It looks like the idea is that if the same customer has multiple copies of the same video checked out, they are expected to return all the duplicate copies at once. Is that right?

app/routes.py
```diff
@@ -1,2 +1,349 @@
...
+
+# List the videos a customer currently has checked out
+def get_video_by_customer_checkout(customer_id):
+
+    customer = Customer.query.get(customer_id)
+
+    if not customer:
+        return({"details":"Invalid data"},400)
+    
+    print(customer)
```
jmaddox19, line 310: In general its best to remove debugging print statements like this before submitting a Pull Request. I assume you probably knew that and just missed this one. Not a big deal :)

jmaddox19 (commented):
Looks good overall! Great work!
Just offered a couple little areas of feedback

PR #32, opened 2021-05-21
Paper Erika Chikere

ErikaChikere:
Will update to include the following optional enhancements: 

1. Customizable query parameters and endpoints


app/models/customer.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+@dataclass
+class Customer(db.Model):
+    id: int 
+    name : str 
+    postal_code: str
+    phone: str
+    registered_at: datetime
+    videos_checked_out_count:int
```
CheezItMan, line 14: I like how you have typing hints here!

app/models/customer.py
```diff
@@ -0,0 +1,44 @@
...
+
+        return result_dict
+
+    @classmethod
+    def from_json(cls,customer_dict): 
+
+        return Customer(name = customer_dict["name"], 
+                postal_code = customer_dict["postal_code"],
+                phone = customer_dict["phone"],
+                videos_checked_out_count = 0)
```
CheezItMan, line 44: In a class method `cls` is this method's class.  So you can technically do this as well.

```suggestion
    @classmethod
    def from_json(cls,customer_dict): 

        return cls(name = customer_dict["name"], 
                postal_code = customer_dict["postal_code"],
                phone = customer_dict["phone"],
                videos_checked_out_count = 0)
```

app/models/customer.py
```diff
@@ -0,0 +1,44 @@
...
+                "registered_at": self.registered_at,
+                "postal_code": self.postal_code,
+                "phone": self.phone,
+                "videos_checked_out_count": 0, 
+
+            }
+
+        return result_dict
+
+    @classmethod
```
CheezItMan, line 38: I love this class method

app/models/rental.py
```diff
@@ -0,0 +1,22 @@
+from flask import current_app
+from app import db
+from dataclasses import dataclass 
+
+
+
+@dataclass
+class Rental(db.Model):
```
CheezItMan, line 8: I suggest that it would be better to move a lot of the "business logic" of your application to the models.  Like having the `check_out` and `check_in` functions in `Rental`.  

This way the `routes.py` file only deals with getting requests and sending back the response.  The logic of checking in and out videos can be done in the models.

app/models/video.py
```diff
@@ -0,0 +1,41 @@
...
+        return result_dict
+
+
+    @classmethod
+    def from_json(cls,video_dict): 
+        return Video(
+            title = video_dict["title"],
+            release_date = video_dict["release_date"],
+            total_inventory = video_dict["total_inventory"],
+            available_inventory = video_dict["total_inventory"])
```
CheezItMan, line 41: Also some good methods in the models.

app/routes.py
```diff
@@ -1,2 +1,284 @@
...
+        return make_response(jsonify(response), 200)
+
+@customer_bp.route("", methods=["POST"])
+def create_a_customer(): 
+    request_body = request.get_json()
+
+        
+    for customer_attribute in ["name","postal_code","phone"]:
+        if customer_attribute not in request_body:
+            return jsonify(f'Missing required: {customer_attribute}'),400
```
CheezItMan, line 33: I like this error response.   However I do think you could create a function to detect and handle errors.  That might help dry up your code.

app/routes.py
```diff
@@ -1,2 +1,284 @@
...
+
+
+    response = {
+             "id": new_customer.id 
+
+               }
+    return make_response(jsonify(response), 201)
+
+@customer_bp.route("/<customer_id>", methods=["GET","PUT","DELETE"])
+def retrieve_one_customer(customer_id):
```
CheezItMan, line 50: You seem to be breaking some routes into individual functions and some into functions like this where the function handles multiple methods.  I would suggest being consistent and either having a function handle multiple methods or have each function handle one method each.  You can dry up code with decorators, if needed.

app/routes.py
```diff
@@ -1,2 +1,284 @@
...
+            return jsonify({}), 400 
+
+    new_video = Video.from_json(request_body)  
+    db.session.add(new_video)
+    db.session.commit()
+
+
+    response = {
+             "id": new_video.id 
+
```
CheezItMan, line 147:
```suggestion
```

app/routes.py
```diff
@@ -1,2 +1,284 @@
...
+# #Routes for Rental 
+@rental_bp.route("/check-out", methods=["POST"])
+def rentals_check_out(): 
+    request_body = request.get_json()
+
+    for rental_attribute in ["customer_id", "video_id"]: 
+        val= request_body[rental_attribute]
+        if not (rental_attribute in request_body) or not isinstance(val,int): 
+            return jsonify({}), 400
```
CheezItMan, line 162: This is another place where you could create a function to validate the input from the request.

app/routes.py
```diff
@@ -1,2 +1,284 @@
...
+
+        
+    
+@video_bp.route("", methods=["POST"])
+def create_a_video(): 
+    request_body = request.get_json()
+
+    for video_attribute in ["title", "total_inventory", "release_date"]: 
+        if not (video_attribute in request_body): 
+            return jsonify({}), 400
```
CheezItMan, line 138: Just another note that you are doing this pattern a lot to check for invalid request body, so it would make a good external function.

app/routes.py
```diff
@@ -1,2 +1,284 @@
...
+        val= request_body[rental_attribute]
+        if not (rental_attribute in request_body) or not isinstance(val,int): 
+            return jsonify({}), 400
+        
+    # foo = ["foo"]
+    # foo["foo"]
+
+    customer = Customer.query.filter_by(id=request_body["customer_id"]).first()
+    video = Video.query.filter_by(id=request_body["video_id"]).first()
+    rental = Rental.query.filter_by(video_id=request_body["video_id"],customer_id=request_body["customer_id"]).first()
```
CheezItMan, line 214: Just another note that you do a lot of queries like this and they would make great methods to put in the model, maybe as class methods. 

That way you would separate the database stuff from handling the requests-responses.

app/routes.py
```diff
@@ -1,2 +1,284 @@
...
+
+  
+    for rental_attribute in ["customer_id", "video_id"]: 
+        val= request_body[rental_attribute]
+        if not (rental_attribute in request_body) or not isinstance(val,int): 
+            return jsonify({}), 400
+        
+    # foo = ["foo"]
+    # foo["foo"]
+
```
CheezItMan, line 211:
```suggestion
```

CheezItMan (commented):
This was really great work Erika!  The code is very solid, so my feedback was more on the high-level organization of the project.  You have all the nuts and bolts of a Flask app well done here.  Just note about how big your `routes.py` file was getting.  Breaking things out and so the `routes.py` file only handles coordinating the request and response and leaves the business logic and database stuff to the model classes would be an organizational improvement.  

Well done!

PR #39, opened 2021-05-21
Paper - Maite A.

MaiteArp:


app/models/customer.py
```diff
@@ -0,0 +1,24 @@
...
+
+    def to_json(self):
+        return  {
+        "id": self.id,
+        "name": self.name,
+        "postal_code": self.postal_code,
+        "phone": self.phone,
+        "registered_at": self.registered_at,
+        "videos_checked_out_count": self.videos_checked_out_count
+        }
```
kaidamasaki, line 22: You can use `rentals` to compute the `videos_checked_out_count` though you need to add a field to `Rental` to keep track of if it's been checked in:

```suggestion
    rentals = db.relationship('Rental', back_populates='customer')

    def to_json(self):
        return  {
        "id": self.id,
        "name": self.name,
        "postal_code": self.postal_code,
        "phone": self.phone,
        "registered_at": self.registered_at,
        "videos_checked_out_count": len([rental for rental in self.rentals if not rental.check_in_date])
        }
```

app/models/rental.py
```diff
@@ -0,0 +1,31 @@
...
+from app import db
+
+
+class Rental(db.Model):
+    id = db.Column(db.Integer, primary_key=True)
+    due_date = db.Column(db.DateTime)
+    check_in_date = db.Column(db.DateTime, nullable=True)
+    customer_id = db.Column(db.Integer, db.ForeignKey('customer.id'))
+    video_id = db.Column(db.Integer, db.ForeignKey('video.id'))
+
```
kaidamasaki, line 12: See note above:

```suggestion
    check_in_date = db.Column(db.DateTime, nullable=True)
```

app/models/video.py
```diff
@@ -0,0 +1,20 @@
...
+    rentals = db.relationship('Rental', back_populates='video') 
+
+    def to_json(self):
+        return {
+        "id": self.id,
+        "title": self.title,
+        "release_date": self.release_date,
+        "total_inventory": self.total_inventory,
+        "available_inventory": self.available_inventory
+        }
```
kaidamasaki, line 20: Similar to above you can compute `available_inventory`:

```suggestion
    available_inventory = db.Column(db.Integer, default=0)
    rentals = db.relationship('Rental', back_populates='video') 

    def to_json(self):
        return {
        "id": self.id,
        "title": self.title,
        "release_date": self.release_date,
        "total_inventory": self.total_inventory,
        "available_inventory": self.total_inventory - len([rental for rental in self.rentals if not rental.check_in_date])
        }
```

kaidamasaki (commented):
Great job!

There were a few little things I pointed out but overall everything looks good.  😃

PR #43, opened 2021-05-21
Paper - Halisa Saipulla

lisa1501:


app/customer_routes.py
```diff
@@ -0,0 +1,243 @@
...
+
+
+@customers_bp.route("", methods=["POST","GET"], strict_slashes = False)
+def create_customers():
+
+    if request.method == "POST":
+        request_body = request.get_json()
+        
+        if "name" not in request_body or "postal_code" not in request_body or "phone" not in request_body: 
+            return ({"details":"Invalid data"},400)
```
CheezItMan: I suggest creating functions to check for invalid fields in the request body.  You could reuse them in several functions.

app/customer_routes.py
```diff
@@ -0,0 +1,243 @@
...
+        due_date=datetime.utcnow()+timedelta(days=7)
+        )
+
+    if video.available_inventory == 0:
+        return({"details":"Invalid data"},400)
+    else:
+        video.available_inventory -= 1
+        customer.videos_checked_out_count += 1
+        db.session.add(rental)
+        db.session.commit()
```
CheezItMan: I strongly suggest pulling this "business logic" out of the `routes.py` file and into the models.  That way the routes only handles coordinating the requests and responses and the models can handle the logic of checking in and out a video.

app/customer_routes.py
```diff
@@ -0,0 +1,243 @@
...
+@rentals_bp.route("/check-in", methods=["POST"], strict_slashes = False)  
+def check_in():
+
+    request_body = request.get_json()
+    
+    if type(request_body["customer_id"]) is not int or "customer_id" not in request_body:
+        return({"details":"Invalid data"},400)
+        
+    if type(request_body["video_id"]) is not int or "video_id" not in request_body:
+        return({"details":"Invalid data"},400)
```
CheezItMan, line 220: Just another area where creating a validation helper method could be useful.

app/models/videos.py
```diff
@@ -0,0 +1,25 @@
+from flask import current_app
+from sqlalchemy.orm import defaultload
+from app.models.rentals import Rental
+from app import db
+
+def default_avail_inv(context): 
+    return context.get_current_parameters()["total_inventory"]
```
CheezItMan: Nice way to add a default for `available_inventory`

app/customer_routes.py
```diff
@@ -0,0 +1,243 @@
+from re import U
```
CheezItMan: I like how it looks like you were trying to break up the `routes.py` file between the types of routes.  It doesn't look like you've done so however.

CheezItMan (commented):
Nice work Halisa, you hit the learning goals here.  I made some suggestions, but they were all high-level on organization rather than specific actions.  Mostly I suggest creating some helper methods for validating requests and pulling out the business logic into the models.  This way your routes file will only handle the specific API request and response logic.  The act of checking in and out videos can be done in the models.

This makes the code a bit more flexible, you could reuse the models in a video store POS system or mobile app for example.  Let me know if you have questions.

PR #44, opened 2021-05-21
Rock- Adelaide Nalley

adiainthesky:


app/models/customer.py
```diff
@@ -0,0 +1,39 @@
+from flask import current_app
+from app import db
+from sqlalchemy import Table, Column, Integer, ForeignKey
+# from sqlalchemy.orm import relationship, back_populate
+# from app.models.video import Video
```
spitsfire, line 5: we can get rid of these. Since we imported `db`, we can already use Column and the other imports like we did on lines 8-14

```suggestion

```

app/models/customer.py
```diff
@@ -0,0 +1,39 @@
...
+
+class Customer(db.Model):
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True) 
+    name = db.Column(db.String)
+    postal_code = db.Column(db.String)
+    phone = db.Column(db.String)
+    created = db.Column(db.DateTime, nullable=True, default=None)
+    videos_checked_out_count = db.Column(db.Integer, default=0, nullable=False)
+    videos = db.relationship("Rental", back_populates="customer")
```
spitsfire, line 14: 👍  I think this is a good solution to stick with. I would consider adding `lazy=True` in the future as it makes it a little easier to query data that has relationships connected to it. `backref` allows you to  relational data from either side. So you can use customer.videos OR, with backref, you can use video.customer to get all the customers connected to the video by rental. 

Based on what I know `backref` only needs to be used in one or the other model. Whereas `back_populates` needs to be used in both tables to connect the relationship more easily

app/models/customer.py
```diff
@@ -0,0 +1,39 @@
...
+    # rentals = db.relationship("Video", secondary="rentals", lazy='subquery',backref=db.backref('customers', lazy=True))
+    
+    #do i need the db. before relationships? what i saw online didnt have it
+    # video_id = db.Column(db.Integer, db.ForeignKey('video.id'), nullable=True) 
+
+    # maybe create a method to see if video.video_checked_out_count:
+    #     else" set to 1. 
+
+
+    def api_response(self):
```
spitsfire, line 29: 👍

app/models/rental.py
```diff
@@ -0,0 +1,17 @@
+from flask import current_app
+from app import db
+from app.models.customer import Customer
+from app.models.video import Video
+from sqlalchemy import Table, Column, Integer, ForeignKey
+# from sqlalchemy.orm import relationship, back_populates
```
spitsfire, line 6:
```suggestion

```

app/models/rental.py
```diff
@@ -0,0 +1,17 @@
...
+
+class Rental(db.Model):
+    #source i got from didnt have db.Column, just Column, integret and foreignkey.  Why?  is it because i dont have rental_id?
+    customer_id = db.Column(db.Integer, db.ForeignKey('customer.id'), primary_key=True)
+    video_id = db.Column(db.Integer, db.ForeignKey('video.id'), primary_key=True)
+    #should i "" Video and Customer?
+    # Customer = relationship("Customer", backref=backref("rentals"))
+    video = db.relationship("Video", back_populates="customers")
+    customer = db.relationship("Customer", back_populates="videos")
```
spitsfire, line 15: you don't need to add these *IF* you use `backref` instead of `back_populates` since they were taken care of in the Customer and Video models themselves, ie `customers = db.relationship("Rental", backref="video")`, so we can just have `customer_id`, `video_id`, and `due_date` in that case.

app/models/video.py
```diff
@@ -0,0 +1,38 @@
+from flask import current_app
+from app import db
+from sqlalchemy import Table, Column, Integer, ForeignKey
+# from sqlalchemy.orm import relationship, back_populate
+# from app.models.customer import Customer
+# from app.models.rental import Rental
```
spitsfire, line 6:
```suggestion

```

app/routes.py
```diff
@@ -1,2 +1,261 @@
-from flask import Blueprint
+from sqlalchemy.orm.base import instance_dict
```
spitsfire, line 1:
```suggestion

```

app/routes.py
```diff
@@ -1,2 +1,261 @@
...
+from datetime import datetime, date, timedelta
+import requests
+import os
+from dotenv import load_dotenv
 
+load_dotenv()
+
+customers_bp = Blueprint("customers", __name__, url_prefix="/customers")
+videos_bp = Blueprint("videos", __name__, url_prefix="/videos")
+rentals_bp = Blueprint("rentals", __name__, url_prefix="/rentals")
```
spitsfire, line 16: 👍 good job separating each model out into its own Blueprint

app/routes.py
```diff
@@ -1,2 +1,261 @@
...
+        return make_response(
+            {"details": "Invalid data"
+            }
+        ), 400
+
+@customers_bp.route("", methods=["GET"])
+def get_customers():
+    request_body = request.get_json()
+    customers = Customer.query.all()
+    customers_response = [customer.api_response() for customer in customers]
```
spitsfire, line 40: 👍 nice list comprehension

spitsfire (commented):
This looks very good, Adelaide! It's pretty consistent, and the code is very readable. I left a few explanations and comments where you had more than one option solution.

PR #47, opened 2021-05-22
Kristin Boyd - Scissors - Video_store_api_development

Kristinboyd:


app/models/customer.py
```diff
@@ -0,0 +1,41 @@
...
+    def to_json_with_id(self):
+        return {
+            "id": self.id
+        }
+
+    def is_int(self):
+        try:
+            return int(self.id)
+        except ValueError:
+            return False
```
jmaddox19, line 37: I see you have this method on the models. I am pretty sure it isn't possible for the id to not be an int, if it's been saved to the db because it's required to by an int by the db, but I would be curious to hear about what motivated you to add these.

app/models/customer.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def current_videos(self):
```
jmaddox19, line 39: This is actually really great! By using data that already exists instead of using an added column (ie. the column `videos_checked_out_count`), you don't run the risk of the data getting out of sync with itself.
Because of that, there shouldn't actually be a need for the column added on line 13

app/models/video.py
```diff
@@ -0,0 +1,34 @@
...
+            "available_inventory": self.get_available_inventory() 
+        }
+
+    def is_int(self):
+        try:
+            return int(self.id)
+        except ValueError:
+            return False
+
+    def get_available_inventory(self):
```
jmaddox19, line 33: This is also really great!

jmaddox19: Again, because you are calculating this value from existing data, the `available_inventory` column can be removed.

jmaddox19 (commented):
Looks really great!
Most people didn't think to calculate the checked_out_count and available_inventory based on the rental rows in the database! Love to see that!

PR #49, opened 2021-05-22
Scissors: Jittania Smith

jittania:
Desperately needs to be refactored but out of time


app/models/customer.py
```diff
@@ -0,0 +1,31 @@
...
+            "id": self.cust_id,
+            "name": self.name,
+            "registered_at": self.registered_at,
+            "postal_code": self.postal_code,
+            "phone": self.phone,  
+            "videos_checked_out_count": self.videos_checked_out_count
+        }
+
+        if rentals_list != None:
+            response_body["rentals"] = rentals_list
```
jmaddox19, line 29: you should be able to use the current_rentals property to create the rentals list instead of requiring it to be passed in here.

jmaddox19: It looks like you didn't end up using that parameter anywhere anyway...but little tidbit for future reference.

app/models/video.py
```diff
@@ -0,0 +1,46 @@
...
+            "title": self.title,
+            "release_date": self.release_date,
+            "total_inventory": self.total_inventory,
+            "available_inventory": self.available_inventory
+        }
+
+        return response_body
+
+
+# ❗️ Does it matter that I didn't give this its own file?
```
jmaddox19, line 27: Python will accept it but people tend to pretty universally prefer each class to be in its own file in most programming languages for readability.

app/models/video.py
```diff
@@ -0,0 +1,46 @@
...
+# ❗️ Does it matter that I didn't give this its own file? 
+class Rental(db.Model): 
+    rental_id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    fk_video_id = db.Column(db.Integer, db.ForeignKey('video.video_id'))
+    fk_customer_id = db.Column(db.Integer, db.ForeignKey('customer.cust_id'))
+    due_date = db.Column(db.DateTime)
+
+    # ❗️ backref here is declaring a 'rental' property for both the Customer and Video classes, but I'm not sure when or if that's being used since I couldn't get either video.current_renters OR customer.current_rentals to work 
+    customer = db.relationship('Customer', backref='rental') 
+    video = db.relationship('Video', backref='rental')
```
jmaddox19, line 36: Because you already have the fk_video_id and fk_customer_id fields, it isn't necessary or typical to add a relationship attribute for those fields here as well. I'm not sure whether there is a way to make use of this in any way because I haven't seen it before.

app/models/video.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+# ❗️ Does it matter that I didn't give this its own file? 
+class Rental(db.Model): 
+    rental_id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    fk_video_id = db.Column(db.Integer, db.ForeignKey('video.video_id'))
+    fk_customer_id = db.Column(db.Integer, db.ForeignKey('customer.cust_id'))
+    due_date = db.Column(db.DateTime)
+
+    # ❗️ backref here is declaring a 'rental' property for both the Customer and Video classes, but I'm not sure when or if that's being used since I couldn't get either video.current_renters OR customer.current_rentals to work
```
jmaddox19, line 34: Those functions look totally appropriate as they are to me, fwiw!

app/routes.py
```diff
@@ -1,2 +1,338 @@
...
+from app.models.video import Video, Rental
+from app.models.customer import Customer
+from flask import request, Blueprint, make_response, jsonify
+
+customers_bp = Blueprint("customers", __name__, url_prefix="/customers")
+videos_bp = Blueprint("videos", __name__, url_prefix="/videos")
+rentals_bp = Blueprint("rentals", __name__, url_prefix="/rentals")
+
+
+# ❗️ SO MUCH REPEAT CODE I'M SO SORRRYYYYYYY - so many opportunities here to DRY up code, just ran out of time
```
jmaddox19, line 13: Totally understand!

jmaddox19: There'll be more opportunities to practice DRYing code

app/routes.py
```diff
@@ -1,2 +1,338 @@
...
+def get_single_customer(customer_id):
+    saved_customer = Customer.query.get_or_404(customer_id)
+    return make_response(saved_customer.convert_to_json(), 200)
+
+
+@customers_bp.route("", methods=["POST"], strict_slashes=False)
+def add_new_customer():
+    request_body = request.get_json()
+
+    # ❗️ Wish I could have refactored this into a helper method (ran out of time) - I use this code so many times
```
jmaddox19, line 57: Great callout! I agree!

app/routes.py
```diff
@@ -1,2 +1,338 @@
...
+    return make_response({ "customer_id": new_rental.fk_customer_id,
+                        "video_id": new_rental.fk_video_id,
+                        "due_date": new_rental.due_date,
+                        "videos_checked_out_count": updated_cust.videos_checked_out_count,
+                        "available_inventory": updated_video.available_inventory
+                    }, 
+                        200)
+
+@rentals_bp.route("/check-in", methods=["POST"], strict_slashes=False)
+# ❗️ I chose to put this in a separate endpoint, wondering if that's correct or if I should have just integrated this code into the check-out endpoint function
```
jmaddox19, line 223: definitely needed as a separate endpoint!

app/routes.py
```diff
@@ -1,2 +1,338 @@
...
+# CUSTOM ENDPOINTS:
+#=============================================================================
+
+@rentals_bp.route("/check-out", methods=["POST"], strict_slashes=False)
+# No route argument necessary for this endpoint
+def check_out_video():
+
+    request_body = request.get_json()
+
+    # ❗️ just awful - I hacked my way through this piecemeal and with more time could have just modified the existing helper function I call here
```
jmaddox19, line 183: I think the code looks pretty good tbh!

app/routes.py
```diff
@@ -1,2 +1,338 @@
...
+    video_id = request_body["video_id"]
+    customer_id = request_body["customer_id"]
+    rental_records = Rental.query.filter_by(fk_video_id=video_id, fk_customer_id=customer_id)
+    for each_rental in rental_records:
+        record_due_date = each_rental.due_date
+        rental_record_id = each_rental.rental_id
+
+    # if the record due date is None:
+    if not record_due_date:
+        return make_response({"details": f"Video with video id {video_id} was already checked in"}, 400)
```
jmaddox19, line 239: great!

app/routes.py
```diff
@@ -1,2 +1,338 @@
...
+                        200)
+
+    
+@customers_bp.route("/<customer_id>/rentals", methods=["GET"], strict_slashes=False)
+def get_customer_rentals(customer_id):
+
+    current_rentals = []
+    current_rentals_response = []
+
+    # ❗️ concerned again about the fact that I'm not really updating the rental table at any point to indicate whether its records are current or not....
```
jmaddox19, line 267: removing the due date does accomplish this, right? Maybe I'm missing something...?

app/routes.py
```diff
@@ -1,2 +1,338 @@
...
+    rental_records = Rental.query.filter_by(fk_customer_id=customer_id)
+
+    # ❗️ I know this can't be the best way to do this (might even be...the worst way), but trying to use `saved_customer.current_rentals` to get some kind of list of current rentals wasn't working in the way I expected, and I just ran out of time
```
jmaddox19, line 270: Huh, bummer you couldn't get that attribute to work! but great that you got this working regardless!

jmaddox19: Again, I really don't think this is bad, even if it is doing a little unnecessary work!

app/routes.py
```diff
@@ -1,2 +1,338 @@
...
+        rental_cust_id = each_rental.fk_customer_id
+        rental_due_date = each_rental.due_date
+        rental_id = each_rental.rental_id
+        renter_info = {"customer_id": rental_cust_id, "due_date": rental_due_date, "rental_id": rental_id}
+        current_renters.append(renter_info)
+
+    # format the response body 
+    for each_renter in current_renters:
+        customer_info = Customer.query.get_or_404(each_renter["customer_id"])
+        rental_info = Rental.query.get_or_404(each_renter["rental_id"])
```
jmaddox19, line 312: the rental info was just retrieved above so there's no need to fetch it again here

jmaddox19: same goes for the similar code in the previous function

jmaddox19 (commented):
Genuinely looks good overall!
I appreciate all the callout notes you left so we can kind of have a conversation about it :) 
Let me know if you have any questions about any of my notes.

PR #50, opened 2021-05-22
Ari - Scissors Class ✂️

arigo7:


app/routes.py
```diff
@@ -1,2 +1,252 @@
...
+        if not video or not customer: 
+            return {"details": "Not Found"}, 404
+        # need to check if video id is in rental
+        elif not "video_id" in request_body or \
+            not "customer_id" in request_body: 
+            return {"details": "Bad Request"}, 400
+        elif not rental:
+            return {"details": "Bad Request"}, 400
+        elif customer.videos_checked_out_count < 1:
+            return {"details": "Bad Request"}, 400
```
beccaelenzil, line 198: Consider providing more details about why the request is bad, such as "That customer doesn't have any Videos checked out", "There is no rental with that customer_id and video_id", "You must provide a customer_id and video_id to look up the rental"

app/routes.py
```diff
@@ -1,2 +1,252 @@
...
+        else:
+        # once all fields valid, will create rental and update due date
+            new_rental = Rental.from_json_to_check_out(request_body)
```
beccaelenzil, line 201: Hrmm.. We do need to create a new rental when we check a video in. Instead, we might decide to update the status of the rental to "checked-in" or delete the rental instance entirely.

app/routes.py
```diff
@@ -1,2 +1,252 @@
...
+        video = Video.query.get(request_body["video_id"])
+        customer = Customer.query.get(request_body["customer_id"])
+
+        if not video or not customer: 
+            return {"details": "Not Found"}, 404
+        elif video.available_inventory < 1: # need to check if inv is lower than
+            return {"details": "Bad Request"}, 400
+        else:
+        # once all fields valid, will create rental and update due date
+            new_rental = Rental.from_json_to_check_out(request_body)
```
beccaelenzil, line 161: Great work encapsulating functionality in a helper method `from_json_to_check_out`. YOu could consider putting additional logic into this method such as the logic below. You would need to pass these instances to the method.

```
customer.videos_checked_out_count += 1            
video.available_inventory -= 1
```

app/routes.py
```diff
@@ -1,2 +1,252 @@
...
+        return {"details": "Bad Request"}, 400
+
+###  GET /customers/<id>/rentals
+### list videos a customer currently has checked out
+@customer_bp.route("/<customer_id>/rentals", methods = ["GET"], \
+    strict_slashes = False)
+def retrieve_videos_checked_out_by_customer(customer_id):
+    customer = Customer.query.get(customer_id) # querying customer by given id
+    if customer == None:
+        return jsonify('Not Found'), 404
```
beccaelenzil, line 221: Above you use this code ` return {"details": "Not Found"}, 404` instead. One isn't better than the other, but being consistent can enhance readability.

app/models/rental.py
```diff
@@ -0,0 +1,54 @@
...
+        input: Takes in a rental instance and a video instance  
+        output: dictionary containing info of instances of video combined 
+        with rental due date details
+        '''
+        return {"release_date": video.release_date.strftime("%Y-%m-%d"),
+                "title": video.title,
+                "due_date": rental.due_date.strftime("%a, %d %b %Y %X %z %Z")}
+
+    @staticmethod
+    def customer_due_date_response(rental, customer):
```
beccaelenzil, line 44: Since this is an instance method, we do not need to pass in the instance of `rental`, we can use `self`

app/models/customer.py
```diff
@@ -0,0 +1,46 @@
...
+class Customer(db.Model):
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    name = db.Column(db.String, nullable=False)
+    postal_code = db.Column(db.Integer, nullable=False)  
+    phone = db.Column(db.String, nullable=True) 
+    # date/time when video was released 
+    registered_at = db.Column(db.DateTime, \
+        default=datetime.now(),
+        nullable=False) 
+    videos_checked_out_count = db.Column(db.Integer, default=0)
```
beccaelenzil, line 14: Instead of `videos_checked_out_count` being a column we can use the rentals to count that.  

```rentals = db.relationship('Rental', back_populates='customer', lazy=True)```

```videos_checked_out_count": len(self.rentals)```

This code wouldn't work exactly as written with your implementation since a rental is created both at check-in and check-out. The code written here assumes a rental is deleted when it was checked it. We might also consider changing the status of the rental to "checked-in" and then writing logic just to find rentals that are checked out.

File: app/models/video.py [hunk 7]
```diff
@@ -0,0 +1,40 @@
+from flask import current_app
+from app import db
+from datetime import datetime
+
+class Video(db.Model):
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)  
+    title = db.Column(db.String) 
+    release_date = db.Column(db.DateTime, nullable=False)
+    total_inventory = db.Column(db.Integer, nullable=False)
+    available_inventory = db.Column(db.Integer, nullable=False)
```
beccaelenzil, line 10: Like with `videos_checked_out_count`, consider how you might calculate this value rather than storing it as an attribute in the database.

beccaelenzil (commented):
Good work working through some tricky logic for this program. You've made good use of instance methods to encapsulate functionality. I've left inline comments on ways to consider refactoring. One place I would encourage you to focus with the next project is being consistent with how code is written to increase readability. Removing comments can also increase readability. I know this sort of refactoring is tricky when there's time pressure. Please let me know if you have any questions.

PR #52, opened 2021-05-22
First Draft for Fridays Deadline

Alli-Oops:
This is what I have completes so far. I will continue working on this over the weekend.


app/routes.py
```diff
@@ -1,2 +1,412 @@
...
+        input_errors.append(blank_name)
+    if 'postal_code' not in request_body.keys() or request_body["postal_code"] == None:
+        blank_postal_code = {"postal_code": "can't be blank"}
+        input_errors.append(blank_postal_code)
+    if 'phone' not in request_body.keys() or request_body["phone"] == None:
+        blank_phone = {"phone": "can't be blank"}
+        input_errors.append(blank_phone)
+
+    if input_errors:
+        return jsonify(errors=input_errors), 400
```
jbieniosek, line 98: This section is very similar to lines 51-62, and is a great candidate for a helper function.

app/routes.py
```diff
@@ -1,2 +1,412 @@
...
+        each_video = v.to_dictionary()
+        videos_response.append(each_video)
+    return jsonify(videos_response), 200
+
+#2############ GET ONE VIDEO by ID - CRUD - READ ###################
+@videos_bp.route("/<video_id>", methods=["GET"])
+def get_video(video_id):
+    if not video_id.isdigit():
+        error_response2 = "is not a number"
+        return jsonify(errors={"video_id": error_response2}), 404
```
jbieniosek, line 144: Great error checking!

app/routes.py
```diff
@@ -1,2 +1,412 @@
...
+
+    video = Video.query.get(video_id)
+    if video:
+        return jsonify(video.to_dictionary()), 200
+    else:
+        error_response1 = "can't be blank"
```
jbieniosek, line 150: Consider the error message here - the video_id can't be blank at this poit since it passed the isdigit test.  "does not exist" would be a more accurate message.

app/routes.py
```diff
@@ -1,2 +1,412 @@
...
+def checkout_to_customer():
+    
+    request_body = request.get_json() 
+    input_errors = []
+    if 'customer_id' not in request_body.keys(): 
+        blank_customer_id = {"customer_id": "can't be blank"}
+        input_errors.append(blank_customer_id)
+    if 'video_id' not in request_body.keys():
+        blank_video_id = {"video_id": "can't be blank"}
+        input_errors.append(blank_video_id)
```
jbieniosek, line 246: This section needs some additional error checking - what are the limitations on the type of data that's valid for customer and video ids?

app/routes.py
```diff
@@ -1,2 +1,412 @@
...
+        input_errors.append(blank_video_id)    
+
+    customer_id = request_body.get("customer_id")
+    video_id = request_body.get("video_id")
+
+    if input_errors:
+        return jsonify(errors=input_errors), 400
+
+    video_thats_rented = Video.query.get(video_id)
+    if video_thats_rented.available_inventory == 0:
```
jbieniosek, line 255: If the video doesn't exist, this line will throw an error.

app/models/customer.py
```diff
@@ -0,0 +1,44 @@
...
+    postal_code: str
+    phone: str
+    videos_checked_out_count: int
+
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    name = db.Column(db.String)
+    registered_at = db.Column(db.DateTime, nullable = True) 
+    postal_code = db.Column(db.String)
+    phone = db.Column(db.String)
+    videos_checked_out_count = db.Column(db.Integer, default=0, nullable = True)
```
jbieniosek, line 19: Could you pull videos_checked_out_count from the rental table rather than storing it directly in the customer table?

app/models/video.py
```diff
@@ -0,0 +1,48 @@
...
+    title: str
+    release_date: datetime
+    total_inventory: int
+    available_inventory: int
+
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    title = db.Column(db.String)
+    release_date = db.Column(db.DateTime)
+    total_inventory = db.Column(db.Integer, default=0)
+    available_inventory = db.Column(db.Integer, default=0)
```
jbieniosek, line 17: Could you pull available_inventory from the rental table rather than storing it directly in this table?

app/routes.py
```diff
@@ -1,2 +1,412 @@
...
+    if 'customer_id' not in request_body.keys(): 
+        blank_customer_id = {"customer_id": "can't be blank"}
+        input_errors.append(blank_customer_id)
+    if input_errors:
+        return jsonify(errors=input_errors), 400
+
+    else: # need to access all the data in customer_id --> not just the integer
+        customer_id = request_body.get("customer_id")
+        update_customer = Customer.query.get(customer_id)
+        customers_rental_ids = update_customer.get("rentals") # accesses the rentals column in the data customer table # get this row with the rental_id
```
jbieniosek, line 306: The rentals attribute of update_customer will be a list of rental model instances.  It can be accessed like any other attribute (update_customer.rentals).

app/routes.py
```diff
@@ -1,2 +1,412 @@
...
+        for rental in customers_rental_ids:
+            if rental.id == update_customer:
+                down_one_count = update_customer.get("videos_checked_out_count") 
+                down_one_count -= 1
```
jbieniosek, line 310: update_customer is an instance of class Customer, while rental.id is will be a number, so this test will always fail (same logic in line 316).  As a result, the variables down_one_count and up_one_count on lines 323 & 324 will never be instantiated.

app/routes.py
```diff
@@ -1,2 +1,412 @@
...
+
+        video_id = request_body.get("video_id")
+        update_video = Video.query.get(video_id)
+        video_rental_ids = update_video.get("rentals") # accesses the rentals column in the data video table
+        for rental in video_rental_ids:
+            if rental.id == update_video:
+                up_one_count = update_video.get("available_inventory")
+                up_one_count += 1
+
```
jbieniosek, line 319: How does the rental table identify that a rental that has ended?

app/routes.py
```diff
@@ -1,2 +1,412 @@
...
+    response = []
+    for customer_id in customers_with_video_list:
+        rental_customer_details = Customer.query.get(customer_id)
+        name = rental_customer_details.name
+        phone = rental_customer_details.phone
+        postal_code = rental_customer_details.postal_code
+        for rental in customer_id.rentals:
+            if rental.video_id == video_id:
+                due_date = rental.due_date # find the rental instance that matches the customer_id and the video_id and get that due date
```
jbieniosek, line 385: Combining these two for loops makes the inner for loop on lines 383-385 unnecessary.  The rental from 374 already has the due date.  Also, there's a small bug on 383, the customer_id is just the id, the rental_customer_details is the class instance with the rentals attribute.

app/routes.py
```diff
@@ -1,2 +1,412 @@
...
+    return jsonify(response), 200
+
+# db.session.query(Video, Customer, Rental).join(Video, Video.id==Rental.video_id).join(Customer, Customer.id==Rental.customer_id).filter(Video.id == who_has_the_vid).all()
+    # ^ result will be an array of tuples. Each tuple will hold 
+    # a Video instance, Customer instance, and a Rental instance 
+        # what_vids_with_customer = ???
+        # videos_with_customer 
+
+######################## GET RENTAL /videos/<id>/rentals CRUD - READ ###################
+@videos_bp.route("/videos/<id>/rentals", methods=["GET"])
```
jbieniosek, line 363: Small bug, this should be ```@videos_bp.route("/<id>/rentals", methods=["GET"])```

jbieniosek:
The Video & Customer CRUD functionality on this project is very solid!  The rental functionality is a little buggy, but most of the logic is there.  Check-out works well except for some error checking.  With Check-in, there's a small piece of the logic that's missing and some AttributeErrors, but other than that it's very close to a working solution.  Overall this project hits all of the learning goals.

PR #57, opened 2021-05-22
Scissors - Melissa N 

melissa-nguyen:


app/models/customer.py
```diff
@@ -0,0 +1,23 @@
...
+from sqlalchemy.orm import relationship
+
+class Customer(db.Model):
+    __tablename__= "customer"
+    id = db.Column(db.Integer, primary_key=True)
+    name = db.Column(db.String)
+    postal_code = db.Column(db.String)
+    phone = db.Column(db.String)
+    registered_at = db.Column(db.DateTime)
+    videos_checked_out_count = db.Column(db.Integer, default=0)
```
beccaelenzil, line 12: Instead of `videos_checked_out_count` being a column we can use the rentals to count that.  

```rentals = db.relationship('Rental', back_populates='customer', lazy=True)```

```videos_checked_out_count": len(self.rentals)```

File: app/models/video.py [hunk 2]
```diff
@@ -0,0 +1,21 @@
...
+from app import db
+from sqlalchemy.orm import relationship
+
+class Video(db.Model):
+    __tablename__= "video"
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    title = db.Column(db.String)
+    release_date = db.Column(db.DateTime)
+    total_inventory = db.Column(db.Integer)
+    available_inventory = db.Column(db.Integer)
```
beccaelenzil, line 11: Like with `Customer`, we can compute `available_inventory` with a relationship:
```rentals = db.relationship('Rental', back_populates='video', lazy=True)```

Then in `to_dict`,  the `available_inventory` is the total inventory minus length of the `rentals`.
```available_inventory": self.total_inventory - len(self.rentals)```

File: app/models/video.py [hunk 3]
```diff
@@ -0,0 +1,21 @@
...
+    customer = relationship("Rental", back_populates="video")
+
+    def to_json(self):
+        return {
+            "id": self.id,
+            "title": self.title,
+            "release_date": self.release_date,
+            "total_inventory": self.total_inventory,
+            # "available_inventory": self.available_inventory
```
beccaelenzil, line 20: Hrmm -- why isn't this included in this dictionary?

app/routes.py
```diff
@@ -1,2 +1,296 @@
...
+        all_videos = Video.query.all()
+
+        videos_response = []
+        for video in all_videos:
+            videos_response.append({
+                "id": video.id,
+                "title": video.title,
+                "release_date": video.release_date,
+                "total_inventory": video.total_inventory,
+                "available_inventory": 0
```
beccaelenzil, line 117: This value should not be hardcoded to `0`, but rather read from the value stored in the instance of the model, or calculated as the `total_inventory` - the length of the rentals for that video.

app/routes.py
```diff
@@ -1,2 +1,296 @@
...
+# endpoint retrieves, updates and deletes records of a specific customer
+@customers_bp.route("/<customer_id>", methods=["GET", "PUT", "DELETE"], strict_slashes=False)
+def specific_customer(customer_id):
+
+    # GET details about specific customer
+
+    customer = Customer.query.get(customer_id)
+
+    if customer is None:
+        return make_response(), 404
```
beccaelenzil, line 66: We could use `get_or_404()` method here 
https://flask-sqlalchemy.palletsprojects.com/en/2.x/api/#flask_sqlalchemy.BaseQuery.get_or_404

app/routes.py
```diff
@@ -1,2 +1,296 @@
...
+        return jsonify(videos_response), 200
+
+    # CREATES (POST) a new VIDEO
+
+    elif request.method == "POST":
+        request_body = request.get_json()
+        if "title" not in request_body or "release_date" not in request_body or "total_inventory" not in request_body:
+            return jsonify({"details": "Invalid data"}), 400
+        else:
+            new_video = Video(
```
beccaelenzil, line 129: Consider removing the `else` to enhance readability.

app/routes.py
```diff
@@ -1,2 +1,296 @@
...
+    # UPDATE (PUT) & return details about specific video
+
+    elif request.method == "PUT":
+        form_data = request.get_json()
+
+        if "title" not in form_data or "release_date" not in form_data or "total_inventory" not in form_data:
+            return jsonify({"details": "Invalid data"}), 400
+
+        else:
+            video.title = form_data["title"]
```
beccaelenzil, line 167: Similar to the comment above, consider removing the else to enhance readability.

app/routes.py
```diff
@@ -1,2 +1,296 @@
...
+        
+            return jsonify({
+                "customer_id": customer_id,
+                "video_id": video_id,
+                "videos_checked_out_count": customer.videos_checked_out_count,
+                "available_inventory": video.available_inventory
+            }), 200
+
+    else:
+        return jsonify({"details": "Invalid data"}), 400
```
beccaelenzil, line 253: For consistency, consider first checking for invalid data, and then getting to the main business of the method if that data is valid.

beccaelenzil (commented):
Good work on this project. Your code is clear and logical, and passes all the test. I encourage you to look back at where you hardcoded `available_inventory` to 0. I suspect this was to get a test to pass, perhaps because the initial database settings had a default for that value. I've left a few other comments on ways you could consider refactoring. Please let me know if you have any questions.

PR #59, opened 2021-05-22
Mariela cruz - scissors 

marielaxcruz:


app/routes.py
```diff
@@ -1,2 +1,231 @@
...
+@rental_bp.route("/check-out", methods=["POST"])
+def checkout_video():
+    request_body = request.get_json()
+
+    # Required Request Body Parameters
+    customer = Customer.query.get(request_body["customer_id"])
+    video = Video.query.get(request_body["video_id"])
+    # check if customer id and video id are int
+    if type(customer) != int or type(video) != int:
+        return jsonify({"details":"Invalid data"}, 404)
```
beccaelenzil: We need to do this type checking before the query, otherwise Customer.query.get will raise an exception if we pass a non integer value.

app/routes.py
```diff
@@ -1,2 +1,231 @@
...
+                    due_date=datetime.now() + timedelta(days=7))
+        # i dk how i can return the customer and video inventory ????
+        ## increase the customer's videos_checked_out_count by one
+        customer.videos_checked_out_count += 1
+        ## decrease the video's available_inventory by one
+        video.available_inventory -= 1
+        
+        db.session.add(new_rental)
+        db.session.add(customer)
+        db.session.add(video)
```
beccaelenzil: We don't need to add `customer` and `video` since these are already in the database, the `db.session.commit()` will take care of persisting any changes.

app/routes.py
beccaelenzil: Also important to note is that customer and video will never be int. We need to check if "customer_id" and "video_id" are numeric values.

app/routes.py
```diff
@@ -1,2 +1,231 @@
...
+
+    # Required Request Body Parameters
+    customer = Customer.query.get(request_body["customer_id"])
+    video = Video.query.get(request_body["video_id"])
+    # check if customer id and video id are int
+    if type(customer) != int or type(video) != int:
+        return jsonify({"details":"Invalid data"}, 404)    
+    # return back detailed errors/ status 404: Not Found if the customer or video does not exist
+    if customer == None or video == None:
+        return jsonify({"details":"Invalid data"}, 404)
```
beccaelenzil: We should not pass 404 into jsonify. There are several different ways to build this response:
`return {"details":"Invalid data"}, 404`
`return jsonify({"details":"Invalid data"}), 404`
`return make_response({"details":"Invalid data"}, 404)`

beccaelenzil: This comment applies to other places where the response code is included as an argument passed to jsonify

app/routes.py
```diff
@@ -1,2 +1,231 @@
...
+        db.session.commit()
+
+        return make_response(new_rental.to_json_rental(), 200)     
+
+# This route checks in a video to a customer, and updates the data in the database
+@rental_bp.route("/check-in", methods=["POST"])
+def customer_check_in():
+    # look up an explanation on what this is doing ????
+    request_body = request.get_json()
+    video_id = checkin_list.get(id)
```
beccaelenzil: I'm not sure what `checkin_list` refers to here

app/routes.py
```diff
@@ -1,2 +1,231 @@
...
+            ## increase the video's available_inventory by one
+            video.available_inventory += 1
+            db.session.commit()
+            return make_response(to_json_rental(), 200)  
+
+
+    else:
+        # return back detailed errors/ status 404: Not Found if the customer or video does not exist
+        if customer == None or video == None:
+            return jsonify({"details":"Invalid data"}, 404)
```
beccaelenzil: We should move this 404 checking to immediately after we query to look for these in the database.

app/models/customer.py
```diff
@@ -0,0 +1,27 @@
...
+    postal_code = db.Column(db.String)
+    phone = db.Column(db.String)
+    registered_at = db.Column(db.DateTime, nullable = True)
+    videos_checked_out_count = db.Column(db.Integer, default = 0) 
+
+    # One to many relationship: one customer has many rentals 
+    customers = db.relationship("Rental", backref="customers", lazy=True)
+
+
+    def to_json(self):
```
beccaelenzil, line 18: Good work putting this functionality into a instance method.

beccaelenzil (commented):
I realize this is incomplete, but here are some comments that may still be helpful.

app/models/video.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+class Video(db.Model):
+    video_id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    title = db.Column(db.String)
+    release_date = db.Column(db.String)
+    total_inventory = db.Column(db.Integer)
+    available_inventory = db.Column(db.Integer)
+    # declare relationship between customer and rental here ?
+    videos = db.relationship("Rental", backref="videos", lazy=True)
```
beccaelenzil, line 13: `rentals = db.relationship("Rental", backref="video")`

app/routes.py
```diff
@@ -1,2 +1,237 @@
...
+    if not isinstance(request_body["customer_id"], int) or not isinstance(request_body["video_id"], int):
+        return {
+            "detail": "Invaid Data. Must be an Integer."
+        }, 404
+    
+    # accessing each video id -- rental and customer id 
+    customer = Customer.query.get(customer_id)
+    video = Video.query.get(video_id)
+
+    for rental in customer.video:
```
beccaelenzil, line 215: To access all the rentals for a particular customer, use your rentals relationship.```for rental in customer.rentals``` -- and the rest of your logic should work out.

app/models/customer.py
```diff
@@ -0,0 +1,27 @@
...
+class Customer(db.Model):
+    customer_id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    name = db.Column(db.String)
+    postal_code = db.Column(db.String)
+    phone = db.Column(db.String)
+    registered_at = db.Column(db.DateTime, nullable = True)
+    videos_checked_out_count = db.Column(db.Integer, default = 0) 
+
+    # One to many relationship: one customer has many rentals 
+    customers = db.relationship("Rental", backref="customers", lazy=True)
```
beccaelenzil, line 15: It looks like this should actually be `rentals`, but you've already dealt with it on the Rental model, so I don't believe you need this here.

app/models/customer.py
```diff
@@ -0,0 +1,27 @@
...
+from datetime import datetime
+from sqlalchemy.orm import backref, relationship
+
+class Customer(db.Model):
+    customer_id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    name = db.Column(db.String)
+    postal_code = db.Column(db.String)
+    phone = db.Column(db.String)
+    registered_at = db.Column(db.DateTime, nullable = True)
+    videos_checked_out_count = db.Column(db.Integer, default = 0)
```
beccaelenzil, line 12: Instead of `videos_checked_out_count` being a column we can use the rentals to count that.  

```rentals = db.relationship('Rental', back_populates='customer', lazy=True)```

```videos_checked_out_count": len(self.rentals)```

File: app/models/video.py [hunk 11]
```diff
@@ -0,0 +1,26 @@
...
+from app import db
+from sqlalchemy.orm import relationship
+
+
+class Video(db.Model):
+    video_id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    title = db.Column(db.String)
+    release_date = db.Column(db.String)
+    total_inventory = db.Column(db.Integer)
+    available_inventory = db.Column(db.Integer)
```
beccaelenzil, line 11: Like with `Customer`, we can compute `available_inventory` with a relationship:
```rentals = db.relationship('Rental', back_populates='video', lazy=True)```

Then in `to_dict`,  the `available_inventory` is the total inventory minus length of the `rentals`.
```available_inventory": self.total_inventory - len(self.rentals)```

beccaelenzil (commented):
You're on your way with the check-in/check-out methods. The logic is there, but there's a bug or two that is keeping the tests from passing. It seems to me that the issue is that you are looping through `customer.videos` instead of `customer.rentals` for the check-in method. Take a look and please let me know if you have any questions. We will check-in about this in our 1:1.

PR #60, opened 2021-05-22
Scissors - Araceli 

aracelim1234:


app/customer_routes.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+@customers_bp.route("", methods=["POST"], strict_slashes=False)
+def new_customer():
+
+    request_body = request.get_json()
+
+    if "name" not in request_body.keys() or "postal_code" not in request_body.keys() or "phone" not in request_body.keys():
+
+        return jsonify({"Error": "Enter info for all fields"}), 400
```
jbieniosek, line 34: 👍

app/customer_routes.py
```diff
@@ -0,0 +1,123 @@
...
+load_dotenv()
+
+@customers_bp.route("", methods=["GET"], strict_slashes=False)
+def get_customers():
+
+    customers = Customer.query.all()
+    customers_respone = []
+
+    if customers is None:
+        return 404
```
jbieniosek, line 19: If there are no customers, query.all() will return an empty list, unlike query.get() which does return None.  This check will always fail, if you want to return an error message in the case of no customers, the conditional needs to check for an empty list or falsey.

app/models/video.py
```diff
@@ -0,0 +1,21 @@
...
+    available_inventory = db.Column(db.Integer, default=0)
+
+    # customers = relationship("Customer", secondary="rental", backref=db.backref("videos"), lazy=True)
+
+    def to_json_video(self):
+        return {
+            "id": self.id,
+            "title": self.title,
+            "release_date": self.release_date,
+            "total_inventory": self.total_inventory
```
jbieniosek, line 20: Small bug here, "available_inventory" should be part of the json output for a video.

app/customer_routes.py
```diff
@@ -0,0 +1,123 @@
...
+    customer = Customer.query.get(customer_id)
+
+    if customer:
+        db.session.delete(customer)
+        db.session.commit()
+
+        return jsonify({"id": customer.id}), 200
+    
+    else:
+        return jsonify({"Error": f"Customer {customer_id} does not exist"}), 404
```
jbieniosek, line 66: Line 49 and this line are doing the same work, but the error messages are different.  Consider pulling this out into a helper function to make it easier to create consistent error messages.  The Flask abort function might be useful in this case, it makes it possible to return a response directly from a helper function.

app/customer_routes.py
```diff
@@ -0,0 +1,123 @@
...
+    videos_list = []
+
+    # if customer exists
+    if customer:
+
+        # create join table 
+        results = db.session.query(Customer, Video, Rental)\
+            .join(Customer, Customer.id==Rental.customer_id)\
+                .join(Video, Video.id==Rental.video_id)\
+                    .filter(Customer.id==id).all()
```
jbieniosek, line 108: 👍

app/rental_routes.py
```diff
@@ -0,0 +1,105 @@
...
+from app.models.video import Video
+from app.models.rental import Rental
+from dotenv import load_dotenv
+from datetime import datetime, timedelta
+
+rentals_bp = Blueprint("rentals", __name__, url_prefix="/rentals")
+load_dotenv()
+
+
+@rentals_bp.route("/check-out", methods=["POST"], strict_slashes=False)
```
jbieniosek, line 13: Breaking up the blueprints into separate files is a great idea!  It makes it much easier to navigate the code :-).

app/rental_routes.py
```diff
@@ -0,0 +1,105 @@
...
+        return jsonify({"Error": "Invalid video input"}), 400
+
+    # get customer_id and create customer instance
+    customer = Customer.query.get(rental.customer_id)
+
+    # get video_id and create video instance
+    video = Video.query.get(rental.video_id)
+
+    # if customer and video exist
+    if customer:
```
jbieniosek, line 35: This only checks to see if the customer exists.  If the video does not exist, line 41 will raise an exception.

app/rental_routes.py
```diff
@@ -0,0 +1,105 @@
...
+    print(rentals[0].customer_id)
+
+    # get customer_id and create customer instance
+    customer = Customer.query.get(rental.customer_id)
+
+    # get video_id and create video instance
+    video = Video.query.get(rental.video_id)
+
+    # if customer exists and is valid 
+    if customer and customer.videos_checked_out >= 0:
```
jbieniosek, line 79: Checking for customer.videos_checked_out >= 0 will allow customer.videos_checked_out to be set to -1.

app/rental_routes.py
```diff
@@ -0,0 +1,105 @@
...
+        
+        # increases videos.total_inventory by 1
+        video.total_inventory += 1
+
+    else:
+        return jsonify({"Error": f"Video {video.id} does not exist"}), 404
+
+    # add it to database and commit
+    db.session.commit()
+
```
jbieniosek, line 98: How are you recording that the rental has ended?  Consider this sequence (customer has 0 videos, video1 has inventory 1, video 2 has inventory 1:
1. A customer checks out 2 videos (customer.videos_checked_out = 2, video1.total_inventory = 0, video2.total_inventory = 0)
2. Check in with customer & video1 (customer.videos_checked_out = 1, video1.total_inventory = 1)
3. Check in with customer & video1:  This should return 400 as this customer does not have this video checked out anymore.  However, it does not - customer.videos_checked_out will be set to 0 and video1.total_inventory will be set to 2, and the API will return a status of 200.  This will set up a situation in the DB where the total inventory has gone up, and the customer has 0 videos checked out, both of which are incorrect.

app/rental_routes.py
```diff
@@ -0,0 +1,105 @@
...
+    video = Video.query.get(rental.video_id)
+
+    # if customer and video exist
+    if customer:
+
+        # increase customer.videos_checked_out by 1
+        customer.videos_checked_out += 1
+    
+        # decrease videos.total_inventory by 1
+        if video.total_inventory > 0:
```
jbieniosek, line 41: The column total_inventory not change when doing rentals, as this is how many copies the video store owns.  There are two ways to check to see if a rental is possible - either use the available_inventory column from models/video.py line 11, or calculate it based on how many rentals have this video id.

app/models/customer.py
```diff
@@ -0,0 +1,25 @@
+from flask import current_app
+from app import db
+from sqlalchemy.orm import relationship, backref
+
+
+class Customer(db.Model):
+    __tablename__ = "customer"
+    id = db.Column(db.Integer, primary_key=True)
+    name = db.Column(db.String)
+    postal_code = db.Column(db.Integer)
```
jbieniosek, line 10: This solution is fine but it causes the code to fail 2 smoke tests (there was inconsistency around the postal_code in the tests).

jbieniosek:
Great work on this project! The code is very clean and readable. There are some bugs with rentals, but the rest of the functionality is solid. This project is green, you hit all of the learning goals.

PR #62, opened 2021-05-22
Paper - Manu C15

kmanuponce:


app/__init__.py
```diff
@@ -1,7 +1,48 @@
...
+    app.config["SQLALCHEMY_TRACK_MODIFICATIONS"] = False
+
+    if test_config is None:
+        app.config["SQLALCHEMY_DATABASE_URI"] = os.environ.get(
+            "SQLALCHEMY_DATABASE_URI")
+# Need lines22-25???
+    # else:
+        app.config["TESTING"] = True
+        app.config["SQLALCHEMY_DATABASE_URI"] = os.environ.get(
+            "SQLALCHEMY_TEST_DATABASE_URI")
```
kaidamasaki, line 25: If you're going to remove an `else` you need to remove all of the lines that are part of its block.

This makes your API always run against the test db.

app/customer_routes.py
```diff
@@ -0,0 +1,95 @@
+from flask import Blueprint, request, jsonify, make_response
+from app import db 
+from app.models.customer import Customer 
+from datetime import date
+from dotenv import load_dotenv
+
+load_dotenv()   #???
```
kaidamasaki, line 7: You already did this in `app/__init__.py` so don't need to repeat it here.

(Also, you don't `os.environ.get` anywhere in this file so it's not necessary anyway.)

app/customer_routes.py
```diff
@@ -0,0 +1,95 @@
...
+
+customer_bp=Blueprint("customers",__name__, url_prefix="/customers")    
+
+# GET /customers
+@customer_bp.route("", methods=["GET","POST"])   #strict_slashes=False??
+def list_customers():   #ok def name?
```
kaidamasaki, line 13: This name implies you're only listing customers in this method.  Since you also create customers here you might want to name it something else.  (For example: `handle_customers`.)

You could also split it into `list_customers` and `create_customer`.

app/customer_routes.py
```diff
@@ -0,0 +1,95 @@
...
+from app.models.customer import Customer 
+from datetime import date
+from dotenv import load_dotenv
+
+load_dotenv()   #???
+
+customer_bp=Blueprint("customers",__name__, url_prefix="/customers")    
+
+# GET /customers
+@customer_bp.route("", methods=["GET","POST"])   #strict_slashes=False??
```
kaidamasaki, line 12: `strict_slashes=False` would let you do `localhost:5000/customers/` in addition to `localhost:5000/customers` (notice the slash at the end).

app/models/customer.py
```diff
@@ -0,0 +1,30 @@
...
+
+# from flask_migrate import Migrate
+
+
+class Customer(db.Model):
+    __tablename__ = "customers"
+
+    customer_id = db.Column(db.Integer, primary_key=True,autoincrement=True)
+    name = db.Column(db.String)    
+    register_at = db.Column(db.DateTime, nullable=True) # nullable=True ???
```
kaidamasaki, line 16: Since you asked, this probably shouldn't be nullable since every customer should have a registration date.

app/models/rental.py
```diff
@@ -0,0 +1,26 @@
...
+# from sqlalchemy import Table, Column, Integer, ForeignKey # Need?
+
+class Rental(db.Model):
+    __tablename__ = "rentals"
+
+    rental_id =db.Column(db.Integer, primary_key=True, autoincrement=True)
+    video_id = db.Column(db.Integer, db.ForeignKey('videos.video_id'),nullable=True)
+    customer_id = db.Column(db.Integer, ForeignKey("customers.customer_id"),primary_key=True)
+    due_date = db.Column(db.DateTime, nullable=True)
+    customer_id = db.Column(db.Integer, ForeignKey("customers.customer_id"),primary_key=True)
```
kaidamasaki, line 15: Instead of importing things from `sqlalchemy` you should just access them through `db`:

```suggestion
from app import db
from sqlalchemy.orm import relationship

class Rental(db.Model):
    __tablename__ = "rentals"

    rental_id =db.Column(db.Integer, primary_key=True, autoincrement=True)
    video_id = db.Column(db.Integer, db.ForeignKey('videos.video_id'),nullable=True)
    customer_id = db.Column(db.Integer, db.ForeignKey("customers.customer_id"),primary_key=True)
    due_date = db.Column(db.DateTime, nullable=True)
    customer_id = db.Column(db.Integer, db.ForeignKey("customers.customer_id"),primary_key=True)
```

You also don't need to import `datetime` since it's different from `db.DateTime`.

app/models/customer.py
```diff
@@ -0,0 +1,30 @@
...
+
+class Customer(db.Model):
+    __tablename__ = "customers"
+
+    customer_id = db.Column(db.Integer, primary_key=True,autoincrement=True)
+    name = db.Column(db.String)    
+    register_at = db.Column(db.DateTime, nullable=True) # nullable=True ??? 
+    postal_code = db.Column(db.String) 
+    phone = db.Column(db.String) 
+    videos_checked_out_count =  db.Column(db.Integer, default=0)
```
kaidamasaki, line 19: Instead of this being a column you can use the rentals to count that.  

```suggestion
    rentals = db.relationship('Rental', back_populates='customer', lazy=True)
```

app/models/customer.py
```diff
@@ -0,0 +1,30 @@
...
+    
+    def get_json(self):
+        return {
+            "id":self.customer_id,
+            "name":self.name,
+            "registered_at":self.register_at,
+            "postal_code":int(self.postal_code),
+            "phone":self.phone,
+            "videos_checked_out_count":self.videos_checked_out_count
```
kaidamasaki, line 28: Once you have the relationship the `videos_checked_out_count` is just the length of the `rentals` that haven't been checked in:

```suggestion
            "videos_checked_out_count": len([rental for rental in self.rentals if not rental.check_in_date])
```

app/models/rental.py
```diff
@@ -0,0 +1,26 @@
...
+
```
kaidamasaki, line 16: You may want to add a check in date to know if the user returned this rental:

```suggestion
    check_in_date = db.Column(db.DateTime, nullable=True)
```

app/models/video.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class Video(db.Model):
+    __tablename__="videos"
+
+    video_id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    title = db.Column(db.String)
+    release_date = db.Column(db.DateTime, nullable=True, default=None) # default=None correct?
+    total_inventory = db.Column(db.Integer, default=0)  
+
```
kaidamasaki, line 16: Like with `Customer` you can compute `available_inventory` with a relationship:

```suggestion
    rentals = db.relationship('Rental', back_populates='video', lazy=True)
```

app/models/video.py
```diff
@@ -0,0 +1,25 @@
...
+    
+
+
+    def video_get_json(self):
+        return {
+            "id":self.video_id,
+            "title":self.title,
+            "release_date":self.release_date,
+            "total_inventory":self.total_inventory,
+            }
```
kaidamasaki, line 25: Like with `Customer` the `available_inventory` is just the total inventory minus length of the `rentals` that haven't been checked in:

```suggestion
            "available_inventory": self.total_inventory - len([rental for rental in self.rentals if not rental.check_in_date])
            }
```

app/rental_routes.py
```diff
@@ -0,0 +1,143 @@
...
+    db.session.add(new_rental)
+    db.session.commit()
+
+
+    return {
+        "customer_id": customer.customer_id,
+        "video_id": video.video_id,
+        "due_date": new_rental.due_date,
+        "videos_checked_out_count": customer.videos_checked_out_count,
+        "available_inventory": video.total_inventory
```
kaidamasaki, line 71: While this makes the tests pass `available_inventory` should not just be `total_inventory` (see earlier comments).

app/rental_routes.py
```diff
@@ -0,0 +1,143 @@
...
+    if not isinstance(request_body["customer_id"], int):
+        return {
+            "details": f"Not found"
+        }, 400
+
+    customer = Customer.query.get(request_body["customer_id"])
+    if customer == None:
+        return {
+            "details": f"Not found"
+        }, 400
```
kaidamasaki, line 35: You can use `get_or_404` to do this more cleanly:

```suggestion
    customer = Customer.query.get_or_404(request_body["customer_id"])
```

app/rental_routes.py
```diff
@@ -0,0 +1,143 @@
...
+    customer = Customer.query.get(request_body["customer_id"])
+    if customer == None:
+        return make_response(f"Customer not found", 404)
+
+    video = Video.query.get(request_body["video_id"])
+    if video == None:
+        return make_response(f"Video not found", 404)
+
+    rentals = Rental.query.filter_by(customer_id=request_body["customer_id"],video_id=request_body["video_id"]).first()
+    if rentals == None:
```
kaidamasaki, line 88: This is only a single rental:

```suggestion
    rental = Rental.query.filter_by(customer_id=request_body["customer_id"],video_id=request_body["video_id"]).first()
    if rental == None:
```

app/rental_routes.py
```diff
@@ -0,0 +1,143 @@
...
+        return {
+            "details": f"Bad Request"
+        }, 400
+
+    customer.videos_checked_out_count -= 1
+    video.total_inventory += 1
+   
+
+    db.session.delete(rentals)
```
kaidamasaki, line 97: Instead of deleting the rental you can mark it as checked in:

```suggestion
    rental.check_in_date = date.today()
```

(This gives you the option of showing checked in rentals if we decide to add new endpoints.)

app/rental_routes.py
```diff
@@ -0,0 +1,143 @@
...
+    
+    rental_list = []
+    for rental in rentals:
+        rented_video = Video.query.get(rental.video_id)
+
+        rental_list.append({
+            "title":rented_video.title,
+            "release_date":rented_video.release_date,
+            "due_date":rental.due_date
+        })
```
kaidamasaki, line 120: If you use relationships you can write this as:

```suggestion
    customer = Customer.query.get_or_404(c_id)
    
    rental_list = []
    for rental in customer.rentals:
        rental_list.append({
            "title": rental.video.title,
            "release_date": rental.video.release_date,
            "due_date": rental.due_date
        })
```

kaidamasaki (commented):
Great job!

I like that you split the routes for the customers and videos into their own files.  It makes it reading it nicer.  :smiley:

I left a few small notes on how to improve but overall everything looks pretty good, well done!

**Note:** It looks like you may have deleted some migrations.

As is you can't cleanly do a `flask db upgrade` on a fresh clone.  

If you're going to delete them (after checking with an instructor or TA) you should probably delete the whole `migrations` folder, drop/re-create the database, and then do a new `flask db init` followed by a `flask db migrate`

PR #64, opened 2021-05-22
Scissors-Lacy-Retro-Video-Store

LacyDraper:


app/models/rental.py
```diff
@@ -0,0 +1,77 @@
...
+                    )
+        
+        db.session.add(new_rental)
+        db.session.commit()
+        customer.increase_checkout_count()#calling Customer instance helper function on instance of Customer
+        video.decrease_inventory()#calling Video helper function on instance of Video
+        return new_rental
+    
+   
+    def to_python_dict(self):
```
jmaddox19, line 45: cool helper method!

jmaddox19 (commented):
Looks great! This is so clean with all the helper functions and methods!

PR #66, opened 2021-05-22
Rock - Abigail C.

ChaAbby:


app/models/customer.py
```diff
@@ -0,0 +1,26 @@
...
+class Customer(db.Model):
+    __tablename__ = 'customer'
+    customer_id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    name = db.Column(db.String)
+    registered_at = db.Column(db.DateTime, default = datetime.now())
+    postal_code = db.Column(db.String)
+    phone = db.Column(db.String)
+    videos_checked_out_count = db.Column(db.Integer, default = 0)
+
+    def to_json(self):
```
spitsfire, line 16: 👍

app/models/rental.py
```diff
@@ -0,0 +1,24 @@
...
+        
+    #     # This method was created so that we do not have to write out the dictionary many times in the routes.py file.
+    #     return {
+    #             "id": self.customer_id,
+    #             "name": self.name,
+    #             "registered_at": self.registered_at,
+    #             "postal_code": self.postal_code,
+    #             "phone": self.phone,
+    #             "videos_checked_out_count": self.videos_checked_out_count
+    #         }
```
spitsfire, line 24: since you don't need this helper method for rentals, make sure you get rid of it before a final submission in the future

```suggestion

```

app/routes.py
```diff
@@ -1,2 +1,256 @@
-from flask import Blueprint
+from flask import Blueprint, request, jsonify, make_response
+from werkzeug.datastructures import Authorization
```
spitsfire, line 2: don't think you are using this below, so we can get rid of imports that aren't be used.

```suggestion

```

app/routes.py
```diff
@@ -1,2 +1,256 @@
...
+
+            db.session.add(customer)
+            db.session.commit()
+            return jsonify(customer.to_json()), 200
+        else:
+            return jsonify({"error": "Bad Request"}), 400
+    elif request.method == "DELETE":
+        db.session.delete(customer)
+        db.session.commit()
+        return { "id" : customer.customer_id }, 200
```
spitsfire, line 41: I think we can organize this a bit to make it more readable. let's think of the method requests (GET, PUT, DELETE) as the main categories, then inside of these we can get more specific. Below is an alternate way to organize with some pseudocode. By no means is it a proper or "correct" solution; simply easier to read I think

```suggestion
    customer = Customer.query.get_or_404(customer_id)
    
    if request.method == "GET":
        return jsonify(customer.to_json()), 200
     elif request.method == "PUT":
         request_body = request.get_json() # I moved this here, because it is only applicable for POST,PUT statements
        # check the request_body for missing values, return 400 and error message
        # else:
            customer.name = request_body["name"]
            customer.postal_code = request_body["postal_code"]
            customer.phone = request_body["phone"]

            db.session.add(customer)
            db.session.commit()
            return jsonify(customer.to_json()), 200
        else:
            return jsonify({"error": "Bad Request"}), 400
            
    elif request.method == "DELETE":
        db.session.delete(customer)
        db.session.commit()
        return { "id" : customer.customer_id }, 200
```

app/routes.py
```diff
@@ -1,2 +1,256 @@
...
+@customers_bp.route("", methods=["GET"])
+def customers_index():
+    customers = Customer.query.all()
+    if customers == None:
+        return [], 200
+    else:
+        customers_response = []
+        for customer in customers:
+            customers_response.append(customer.to_json())
+        return jsonify(customers_response), 200
```
spitsfire, line 52: one nice thing about `query.all()` is that if no data is returned, it will still create an empty list for us, so the `if` statement won't ever be triggered. Maybe we could do something like this:

```suggestion
    customers = Customer.query.all()
    if len(customers) < 1:
        return customers, 200
    else:
        customers_response = []
        for customer in customers:
            customers_response.append(customer.to_json())
        return jsonify(customers_response), 200
```

app/routes.py
```diff
@@ -1,2 +1,256 @@
...
+
+        return jsonify({"id": new_customer.customer_id}), 201
+    except KeyError:
+        return jsonify({
+            "details": "Invalid data"}), 400
+
+# ================================== Video ========================================================
+
+@videos_bp.route("/<video_id>", methods = ["GET", "PUT", "DELETE"])
+def get_single_video(video_id):
```
spitsfire, line 74: same as above, I think we could organize these conditional statements a little better

app/routes.py
```diff
@@ -1,2 +1,256 @@
...
+@videos_bp.route("", methods=["GET"])
+def videos_index():
+    videos = Video.query.all()
+    if videos == None:
+        return [], 200
+    else:
+        videos_response = []
+        for video in videos:
+            videos_response.append(video.to_json_video())
+        return jsonify(videos_response), 200
```
spitsfire, line 113: same as above! We should change the comparison a bit

app/routes.py
```diff
@@ -1,2 +1,256 @@
...
+#         return jsonify(videos_response), 200
+
+
+@rentals_bp.route("/check-out", methods = ["POST"])
+def rentals_check_out():
+
+    if request.method == "POST":
+        request_body = request.get_json()
+        request_body_customer = request_body.get('customer_id')
+        request_body_video = request_body.get('video_id')
```
spitsfire, line 182: right here we want to query the specific customer and video. This will allow line 187 to work. Right now, we are grabbing the value of the key "customer_id", which is a number, not the entire customer.

```suggestion
        request_body_customer = Customer.query.get(request_body['customer_id'])
        request_body_video = Video.query.get(request_body['video_id'])
```

app/routes.py
```diff
@@ -1,2 +1,256 @@
...
+        if request_body_customer == None:
+            return jsonify({"details":"customer does not exist"}), 404
+        elif request_body_video == None:
+            return jsonify({"details": "video does not exist"}), 404
+        elif request_body_video.get('available_inventory') == 0:
+            return jsonify({"details": "There is not available inventory"}), 400
+        else:
+            checked_out = Rental(
```
spitsfire, line 190: now here we need to remember to add in the `customer_id` and `video_id` to make the connection between the two, so those should be added to the Rental instance

app/routes.py
```diff
@@ -1,2 +1,256 @@
...
+                            "videos_checked_out_count": checked_out.videos_checked_out_count,
+                            "available_inventory": checked_out.update_inventory
+                            }), 200
+
+@rentals_bp.route("/check-in", methods = ["POST"])
+def rentals_check_in():
+    if request.method == "POST":
+        request_body = request.get_json()
+        request_body_customer = request_body.get('customer_id')
+        request_body_video = request_body.get('video_id')
```
spitsfire, line 210: let's do the same thing here as above

app/routes.py
```diff
@@ -1,2 +1,256 @@
...
+    customers = Customer.query.all()
+    if request.customer_id is None:
+        return jsonify({"details": "cutomer does not exist"}), 404
+    elif customers == None:
+        return [], 200
+    else:
+        customers_response = []
+        for customer in customers:
+            customers_response.append(customer.to_json())
+        return jsonify(customers_response), 200
```
spitsfire, line 243: we can do this a couple of ways, but we might want to query the rentals that are connecting to the *one* customer by the `customer_id`. From there we can find out what videos the customer rented out by the `video_id` connected to the rentals and query each video.

Then return the list of videos the customer rented

app/routes.py
```diff
@@ -1,2 +1,256 @@
...
+    videos = Video.query.all()
+    if request.video_id is None:
+        return jsonify({"details": "video does not exist"}), 404
+    elif videos == None:
+        return [], 200
+    else:
+        videos_response = []
+        for video in videos:
+            videos_response.append(video.to_json_video())
+        return jsonify(videos_response), 200
```
spitsfire, line 256: this one, we may want to start in Rental again. Query all rentals based on the `video_id`, then find each customer that rented out the video and return that list.

spitsfire (commented):
Great job with Wave 1, Abby! I've added a few suggestions on what can be done with Wave 2 (though don't feel pressured to do this right now! just something to consider)

PR #69, opened 2021-05-22
Scissors - Aida R.

adia-9:


app/models/customer.py
```diff
@@ -0,0 +1,23 @@
...
+from sqlalchemy.orm import relationship
+from sqlalchemy.ext.declarative import declarative_base
+
+class Customer(db.Model):
+    customer_id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    name = db.Column(db.String)
+    postal_code = db.Column(db.String)
+    phone = db.Column(db.String)
+    register_at = db.Column(db.DateTime)
+    videos_checked_out_count = db.Column(db.Integer, default=0)
```
jbieniosek, line 12: Could you pull videos_checked_out_count from the rental table rather than storing it directly in the customer table?

app/models/video.py
```diff
@@ -0,0 +1,20 @@
...
+from app import db
+from sqlalchemy.orm import relationship
+from sqlalchemy.ext.declarative import declarative_base
+
+class Video(db.Model):
+    video_id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    title = db.Column(db.String)
+    release_date = db.Column(db.DateTime)
+    total_inventory = db.Column(db.Integer, default=0)
+    available_inventory = db.Column(db.Integer, default=total_inventory)
```
jbieniosek, line 11: Could you pull available_inventory from the rental table rather than storing it directly in this table?

app/routes.py
```diff
@@ -1,2 +1,253 @@
...
+    db.session.add(new_customer)
+    db.session.commit()
+    return make_response({"id": new_customer.customer_id}, 201)
+
+@customer_bp.route("", methods = ["GET"], strict_slashes = False)
+def get_all_customers():
+    customers = Customer.query.all()
+    customer_list = []
+    for customer in customers:
+        customer_list.append(customer.to_json())
```
jbieniosek, line 42: This works well, but is a great candidate for a list comprehension:
```
customer_list = [customer.to_json() for customer in customers]
```

app/routes.py
```diff
@@ -1,2 +1,253 @@
...
+    
+    return make_response(jsonify(customer_list), 200)
+
+@customer_bp.route("/<customer_id>", methods = ["GET"], strict_slashes = False)
+def get_one_customer(customer_id):
+    customer = Customer.query.get(customer_id)
+    if customer:
+        return make_response(customer.to_json(), 200)
+
+    return customer_not_found(customer_id)
```
jbieniosek, line 52: Nice helper function!

app/routes.py
```diff
@@ -1,2 +1,253 @@
...
+    customer = Customer.query.get(customer_id)
+    video = Video.query.get(video_id)
+
+    if not customer:
+        return customer_not_found(customer_id)
+    if not video:
+        return video_not_found(video_id)
+
+    if video.available_inventory == 0:
+        return make_response({"details": "No inventory available"}, 400)
```
jbieniosek, line 159: 👍

app/routes.py
```diff
@@ -1,2 +1,253 @@
...
+    
+    if customer.videos_checked_out_count == 0:
+        return make_response({"details": "All videos are returned"}, 400)
+
+    video.available_inventory += 1
+    customer.videos_checked_out_count -= 1
+
+    new_rental = Rental(customer_id = request_body["customer_id"], video_id = request_body["video_id"])
+    db.session.add(new_rental)
+    db.session.commit()
```
jbieniosek, line 202: This creates a new entry in rental with a null due date.  It does not remove the existing rental.  The rental will still show up on /customers/<id>/rentals or /vidoes/<id>/rentals.  How are you recording that the rental has ended?

app/routes.py
```diff
@@ -1,2 +1,253 @@
...
+        "videos_checked_out_count": customer.videos_checked_out_count,
+        "available_inventory": video.available_inventory
+    },200)
+
+@video_bp.route("/<video_id>/rentals", methods=["GET"])
+def rental_video_by_id(video_id):
+    rental_results = db.session.query(Customer, Video, Rental).join(
+        Video, Video.video_id==Rental.video_id).join(
+        Customer, Customer.customer_id==Rental.customer_id).filter(
+        Video.video_id==video_id).all()
```
jbieniosek, line 216: 👍

jbieniosek:
Great work on this project! The code is very clean and readable.  There's a small bug in check-in, but the rest of the functionality is solid.  This project is green, you hit all of the learning goals.

app/models/video.py
jbieniosek, line 11: As an aside, I don't think setting the default to total_inventory will work - it never gets called, because the only time a new video is created (routes.py, line 91) a value is passed in for available_inventory, so there is never an attempt to use the total_inventory as a default.

PR #72, opened 2021-05-22
Rock-Briyana Haywood

brijhaywood:


app/routes.py
```diff
@@ -1,2 +1,212 @@
...
+from flask import request, Blueprint, make_response, jsonify
+from datetime import datetime
+import requests
 
+
+customers_bp = Blueprint("customers", __name__, url_prefix="/customers")
+videos_bp = Blueprint("videos", __name__, url_prefix="/videos")
+rentals_bp = Blueprint("rentals", __name__, url_prefix="/rentals")
+
+def customer_response_helper(customer):
```
spitsfire, line 14: 👍 yay helper functions! one thing to think about in the future is any functions like this should be put in the models. the routes should be very bare bones. Routes should only query the database, call model methods, and direct the return statement. Everything else, thinkaboutt putting into models or a separate helper file!

app/routes.py
```diff
@@ -1,2 +1,212 @@
...
+
+@customers_bp.route("", methods = ["GET"])
+def get_customers():
+    customers = Customer.query.all()
+    customers_response = []
+    if customers is None:
+        return make_response([])
+    for customer in customers:
+        customers_response.append(customer_response_helper(customer))
+    return jsonify(customers_response)
```
spitsfire, line 50: since `query.all()` returns an empty list when no data found, I'm wondering if this if statement will ever be tripped.  Maybe we could combine these a little more:

```suggestion
def get_customers():
    customers = Customer.query.all()
    customers_response = []
    if len(customers) < 1:
        for customer in customers:
            customers_response.append(customer_response_helper(customer))
    return jsonify(customers_response)
```

app/routes.py
```diff
@@ -1,2 +1,212 @@
...
+
+@customers_bp.route("/<customer_id>", methods = ["GET"])
+def get_one_customer(customer_id):
+    customer = Customer.query.get_or_404(customer_id)
```
spitsfire, line 54: 👍 great to see `get_or_404()`

app/routes.py
```diff
@@ -1,2 +1,212 @@
...
+            "name": new_customer.name,
+            "postal_code": new_customer.postal_code,
+            "phone": new_customer.phone_number
+        }, 201)
+
+@customers_bp.route("/<customer_id>", methods = ["PUT"])
+def update_customer(customer_id):
+    customer = Customer.query.get(customer_id)
+    if customer is None:
+        return make_response({"details": "Customer not found"}, 404)
```
spitsfire, line 77: here we could also use `get_or_404()`

```suggestion
    customer = Customer.query.get_or_404(customer_id)
```

app/routes.py
```diff
@@ -1,2 +1,212 @@
...
+        return make_response({"details": "Invalid data"}, 400)
+    customer.name = form_data["name"]
+    customer.postal_code = form_data["postal_code"]
+    customer.phone_number = form_data["phone"]
+    db.session.commit()
+    return customer_response_helper(customer)
+
+@customers_bp.route("/<customer_id>", methods = ["DELETE"])
+def delete_customer(customer_id):
+    customer = Customer.query.get(customer_id)
```
spitsfire, line 89: here we could use `get_or_404()` too

app/routes.py
```diff
@@ -1,2 +1,212 @@
...
+
+@videos_bp.route("", methods = ["GET"])
+def get_videos():
+    videos = Video.query.all()
+    videos_response = []
+    if videos is None:
+        return []
+    for video in videos:
+        videos_response.append(video_response_helper(video))
+    return jsonify(videos_response)
```
spitsfire, line 104: same here, we could combine these like lines 43-50

app/routes.py
```diff
@@ -1,2 +1,212 @@
...
+        new_video = Video(title=request_body["title"], release_date=request_body["release_date"], total_inventory=request_body["total_inventory"])
+        db.session.add(new_video)
+        db.session.commit()
+        return make_response({
+            "id": new_video.video_id
+        }, 201)
+
+@videos_bp.route("/<video_id>", methods = ["PUT"])
+def update_videos(video_id):
+    video = Video.query.get(video_id)
```
spitsfire, line 126: here we could use `get_or_404()` too

app/routes.py
```diff
@@ -1,2 +1,212 @@
...
+    else:
+        video.title = form_data["title"]
+        video.release_date = form_data["release_date"]
+        video.total_inventory = form_data["total_inventory"]
+        db.session.commit()
+        return video_response_helper(video)
+
+@videos_bp.route("/<video_id>", methods = ["DELETE"])
+def delete_video(video_id):
+    video = Video.query.get(video_id)
```
spitsfire, line 141: here we could use `get_or_404()` too

app/routes.py
```diff
@@ -1,2 +1,212 @@
...
+    elif type(request_body["customer_id"]) is not int :
+        return make_response({"details": "Invalid customer_id"}, 400)
+    elif type(request_body["video_id"]) is not int :
+        return make_response({"details": "Invalid video_id"}, 400)
+    new_rental = Rental(customer_id=request_body["customer_id"], video_id=request_body["video_id"])
+    customer = Customer.query.get(new_rental.customer_id)
+
+    video = Video.query.get(new_rental.video_id)
+    # customer.videos_checked_out_count += 1
+    # video.available_inventory -=1
```
spitsfire, line 163: Hmm any reason you commented these out? We need to account for increasing/decreasing inventory and checkout count for Wave 2

```suggestion

```

app/routes.py
```diff
@@ -1,2 +1,212 @@
...
+    customer = Customer.query.get(customer_id)
+    rental = Rental.query.filter_by(customer_id= customer.customer_id)
+    list_of_rentals = []
+    for each_rental in rental:
+        video = Video.query.get(each_rental.video_id)
+        list_of_rentals.append({
+            "release_date": video.release_date,
+            "title": video.title,
+            "due_date": each_rental.due_date,    
+            })
```
spitsfire, line 195: we could turn this into a helper function

app/routes.py
```diff
@@ -1,2 +1,212 @@
...
+    rental = Rental.query.filter_by(video_id= video.video_id)
+    list_of_rentals = []
+    for each_rental in rental:
+        customer = Customer.query.get(each_rental.customer_id)
+        list_of_rentals.append({
+            "due_date": each_rental.due_date, 
+            "name": customer.name,
+            "phone": customer.phone_number,
+            "postal_code": customer.postal_code   
+            })
```
spitsfire, line 211: this could also be turned into a helper function

spitsfire (commented):
Briyana, great job! One thing to think about is consistency! Sometimes you use `get_or_404()`, sometimes you don't. Sometimes you use `make_response`, sometimes you return the data by itself. And that all works! But APIs work best when they are predictable and give consistent return statements and such.

Otherwise, this is very nice! You are missing some Wave 2 smoke tests, but the main CRUD operations are working well

PR #77, opened 2021-05-22
rock - mana

mahnnotman:


app/routes.py
```diff
@@ -1,2 +1,205 @@
...
+        for customer in customers:
+            customers_response.append({
+                "id": customer.customer_id,
+                "name": customer.name,
+                "registered_at": customer.register_at,
+                "postal_code": customer.postal_code,
+                "phone": customer.phone,
+                "videos_checked_out_count": 0
+            })
+        return jsonify(customers_response), 200
```
spitsfire, line 36: 👍 this works perfectly! One thing to consider, though, is that the return statement is exactly the same, so let's save a line. Also, `query.all()` will always return a `[]` whether it is full of data or not, so we don't need to check if it's empty at first.  Either way, `customer_response` will be returned now:

```suggestion
    customers = Customer.query.all()
    customers_response = []
    
    for customer in customers:
        customers_response.append({
            "id": customer.customer_id,
            "name": customer.name,
            "registered_at": customer.register_at,
            "postal_code": customer.postal_code,
            "phone": customer.phone,
            "videos_checked_out_count": 0
        })
    return jsonify(customers_response), 200
```

app/routes.py
```diff
@@ -1,2 +1,205 @@
...
+
+
+@customers_bp.route("/<customer_id>", methods=["GET"], strict_slashes=False)
+def handle_single_customer(customer_id):
+
+    customer = Customer.query.get(customer_id)
+
+    if customer is None:
+        return jsonify(f"customer {customer_id} doesn't exist."), 404
```
spitsfire, line 45: we can do this all in one line:

```suggestion
    customer = Customer.query.get_or_404(customer_id)
```

app/routes.py
```diff
@@ -1,2 +1,205 @@
...
+
+    db.session.add(new_customer)
+    db.session.commit()
+
+    return jsonify({"id": new_customer.customer_id}), 201
+
+
+@customers_bp.route("/<customer_id>", methods=["PUT"], strict_slashes=False)
+def update_single_customer(customer_id):
+    customer = Customer.query.get(customer_id)
```
spitsfire, line 76: here we can use `get_or_404()` too!

app/routes.py
```diff
@@ -1,2 +1,205 @@
...
+
+    else:
+        for video in videos:
+            videos_response.append({
+                "id": video.video_id,
+                "title": video.title,
+                "release_date": video.release_date,
+                "total_inventory": video.total_inventory,
+                "available_inventory": 0})
+        return jsonify(videos_response), 200
```
spitsfire, line 131: here we can also shorten this up like on lines 20-36

app/routes.py
```diff
@@ -1,2 +1,205 @@
...
+
+
+@videos_bp.route("/<video_id>", methods=["GET"], strict_slashes=False)
+def handle_single_video(video_id):
+
+    video = Video.query.get(video_id)
```
spitsfire, line 137: here we can use `get_or_404()` too!

app/routes.py
```diff
@@ -1,2 +1,205 @@
...
+    return jsonify({"id": new_video.video_id,
+        "title": new_video.title,
+        "release_date": new_video.release_date,
+        "total_inventory": new_video.total_inventory,
+        "available_inventory": 0}), 201
+
+
+@videos_bp.route("/<video_id>", methods=["PUT"], strict_slashes=False)
+def update_single_video(video_id):
+    video = Video.query.get(video_id)
```
spitsfire, line 173: here we can use `get_or_404()` too!

app/routes.py
```diff
@@ -1,2 +1,205 @@
...
+    return jsonify({"id": video.video_id,
+        "title": video.title,
+        "release_date": video.release_date,
+        "total_inventory": video.total_inventory,
+        "available_inventory": 0}), 200
+
+
+@videos_bp.route("/<video_id>", methods=["DELETE"], strict_slashes=False)
+def delete_single_video(video_id):
+    video = Video.query.get(video_id)
```
spitsfire, line 197: here we can use `get_or_404()` too!

spitsfire (commented):
heck yeah, Mana! This was very nicely structured! Even if you didn't finish up Wave 2, that's ok! Your main CRUD operations look very good *and* you remembered to stay consistent with your return statements. They are all `return jsonify()` woo hoo!!

We could also turn some of the for loops that reformat our data into model methods inside the model files. Something to consider in the future! Otherwise, you did very nice!

PR #78, opened 2021-05-22
Raquel Mena // Scissors

Raque-lita:
30 minutes late to finish passing tests


jmaddox19:
It looks like this project is missing a .gitignore file so unfortunately the diff is so big I can't do inline comments.

Love the separate methods for checkout and checkin in the model!

Line 43 in the rental model is a little awkward because returning json from the method won't actually return that to the user, since it isn't a route function. Let me know if that doesn't make sense and we can talk about in in our 1:1 next week!

In the checkin route, there's a gap where the rental table in the database isn't altered in any way to signal that the rental has been checked in, so the customer can potentially check in the same rental over and over. I see that the code checks whether the video.checked_out_count is 0, but this won't stop the same customer from checking in the video they don't have if another customer has checked out the same video. Let's talk about this in our next 1:1 as well, if it isn't super clear!

PR #79, opened 2021-05-22
Lindsey - retro-video-store

lileekoi:
Wave 2 incomplete. Will resubmit over the weekend


jmaddox19:
I have pulled in the code to run the tests and see what you're saying about it being incomplete. As is, this is Yellow, but the code that is there looks good. We won't ask you to revisit this given how much of it you did get done, but if you were to have some extra energy and wanted to come back to this, I would suggest focusing on handling the edge cases for check-in and check-out and probably not bothering with the final /customers/rentals and /videos/rentals endpoints.

There are opportunities for additional functions to DRY up the code but that is something we only recommend focusing on if you have additional time. One example is the repeated code that looks like:
```
for prop in required_properties:
        if prop not in request_body:
            return make_response({"details": "Invalid data"}, 400)
```
This would be a great candidate for a helper function.

For future reference, something went wrong here that made this difficult to review. The venv directory and the .env file are in the original .gitignore file so they should have been ignored by git but I wonder if you accidentally deleted the .gitignore file at some point or something. the venv directory actually has thousands of files in it so its causing github to crash when I try to look at the files.

PR #82, opened 2021-05-23
Paper - Ekaterina Malakhova

kmalakhova:


app/models/customer.py
```diff
@@ -0,0 +1,20 @@
...
+
+class Customer(db.Model):
+    __tablename__ = "customer"
+    customer_id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    name = db.Column(db.String)
+    registered_at = db.Column(db.DateTime(), nullable=True)
+    postal_code = db.Column(db.Integer)
+    phone = db.Column(db.String)
+    videos_checked_out_count = db.Column(db.Integer, default=0)
+
```
CheezItMan, line 11: I would also consider adding relationships as it can make it easier to access rentals by customers etc.

```python
rentals = db.relationship('Rental',backref = 'customer')
```

You can do this in each of your model classes.

app/models/rental.py
```diff
@@ -0,0 +1,16 @@
+from app import db
+
+class Rental(db.Model):
+    __tablename__ = "rental"
+    rental_id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    customer_id = db.Column("customer", db.Integer, db.ForeignKey("customer.customer_id"), primary_key=True)
+    video_id = db.Column("video", db.Integer, db.ForeignKey("video.video_id"), primary_key=True)
+    due_date = db.Column(db.DateTime)
+
```
CheezItMan, line 9: I also suggest that you add some business logic methods for handling checking out and checking in.  This way you can have some separation of concerns between the routes and the models.

It dryes up the the routes file and makes your code more flexible so that you could move the models to a different type of app (like a mobile app, or desktop app etc), and reuse the code.  That's harder to do if the business logic is mixed in with the specific API routes-requests content.

app/routes.py
```diff
@@ -1,2 +1,264 @@
...
+
+def bad_request():
+        return jsonify({400: "Bad Request"}), 400
+
+def not_found():
+        return jsonify({404: "Not Found"}), 404
+
+def error_response(code=400):
+    return {"details": "Invalid data"}, code
+
```
CheezItMan, line 25: Nice helper functions.

app/routes.py
```diff
@@ -1,2 +1,264 @@
...
+        customer_id=customer_id, 
+        video_id=video_id,
+        due_date = date.today() + timedelta(days=7)
+        )
+    
+    customer.videos_checked_out_count += 1
+    video.available_inventory -= 1
+    
+    db.session.add(rental)
+    db.session.commit()
```
CheezItMan, line 175: This is the kind of logic you could move into the models to handle the checking in and checking out logic.

app/routes.py
```diff
@@ -1,2 +1,264 @@
...
+        video_id=video.video_id).one_or_none()
+
+    if rental is None:
+        return bad_request()   
+
+    video.available_inventory += 1
+    customer.videos_checked_out_count -= 1
+
+    db.session.delete(rental)
+    db.session.commit()
```
CheezItMan, line 212: Similarly this would make a great thing to move into a method in the models.

app/routes.py
```diff
@@ -1,2 +1,264 @@
...
+
+@customers_bp.route("<customer_id>/rentals", methods=["GET"])
+def get_customer_rentals(customer_id):
+    """List the videos a customer currently has checked out"""
+    customer = Customer.query.get(customer_id)
+
+    if customer is None:
+        return not_found()
+    
+    rentals = Rental.query.filter_by(customer_id=customer.customer_id)
```
CheezItMan, line 229: If you use the relationship you don't need to use the whole filter thingy.

You can instead do something like:

`rentals = customer.rentals`

Which helps get the database "junk" out of the routes file too.

venv~/bin/Activate.ps1
```diff
@@ -0,0 +1,241 @@
+<#
```
CheezItMan, line 1: I really suggest adding `venv` to your `.gitignore` file.  No one wants to see this in a PR.

CheezItMan (commented):
Nice work Kate, you hit the learning goals here.  I made a few suggestions, mostly on organization of code and using database relationships.  The highlights include putting "business logic" into the models and move it outside of the routes and using relationships to avoid having to do so much `.query.filter_by` etc code. 

Overall this is quite well done.  Nice work!

PR #83, opened 2021-05-24
Paper: Saejin Tanguay

saebaebae:


app/models/customer.py
```diff
@@ -0,0 +1,28 @@
...
+    customer_id = db.Column(db.Integer, primary_key=True)
+    name = db.Column(db.String, nullable=False)
+    postal_code = db.Column(db.String, nullable=False)
+    phone = db.Column(db.String, nullable=False)
+    registered_at = db.Column(db.DateTime, nullable = True)
+    videos_checked_out_count = db.Column(db.Integer, default=0)
+    # videos = db.relationship("Video", secondary="rental", lazy=True)
+
+
+    def json_response(self):
```
CheezItMan, line 18: Good model method

app/models/customer.py
```diff
@@ -0,0 +1,28 @@
...
+from app import db
+# from app.models.video import Video
+# from app.models.rental import Rental
+# from flask_sqlalchemy import SQLAlchemy
+
+
+class Customer(db.Model):
+    customer_id = db.Column(db.Integer, primary_key=True)
+    name = db.Column(db.String, nullable=False)
+    postal_code = db.Column(db.String, nullable=False)
```
CheezItMan, line 11: I think that's fine as a string

app/models/customer.py
```diff
@@ -0,0 +1,28 @@
+from flask import current_app
+from app import db
+# from app.models.video import Video
+# from app.models.rental import Rental
+# from flask_sqlalchemy import SQLAlchemy
```
CheezItMan, line 5:
```suggestion
```

app/models/rental.py
```diff
@@ -0,0 +1,10 @@
+from flask import current_app
+from app import db
+# from app.models.video import Video
+# from app.models.customer import Customer
+
+class Rental(db.Model):
```
CheezItMan, line 6: Note:  It would make a great deal of sense to add more methods to the model classes to handle the "business logic" of the application.  Things like checking out a video and checking in a video.

This would accomplish a few things:

1.  It would separate the specific application logic (into the models) away from the part getting the requests and sending back the responses (routes.py).
1.  It would lead to dryer code so your routes functions are not too long.
1.  It would make the code more reusable by decouping the specific API logic from the video store POS (Point of Service) logic.

app/models/video.py
```diff
@@ -0,0 +1,23 @@
+from flask import current_app
+from app import db
+
+def mydefault(context):
+        return context.get_current_parameters()['total_inventory']
```
CheezItMan, line 5: Interesting way to get a default value for `available_inventory`

app/routes.py
```diff
@@ -1,2 +1,317 @@
...
+from flask import request, Blueprint, make_response, jsonify
+import os
+import requests
+from datetime import datetime, timedelta
 
+now=datetime.now()
+
+customer_bp = Blueprint("customer", __name__, url_prefix="/customers")
+video_bp = Blueprint("video", __name__, url_prefix="/videos")
+rental_bp = Blueprint("rental", __name__, url_prefix="/rentals")
```
CheezItMan, line 14: 👍

app/routes.py
```diff
@@ -1,2 +1,317 @@
...
+from app import db
+from app.models.video import Video
+from app.models.customer import Customer
+from app.models.rental import Rental
+from flask import request, Blueprint, make_response, jsonify
+import os
+import requests
+from datetime import datetime, timedelta
 
+now=datetime.now()
```
CheezItMan, line 10: Just note that `now` will be the time the application started.  So if the API keeps running for 3-5 days, your datetime here will become increasingly "outdated"  🤣 

In other words `now` won't be updated with each request, but set when you run `flask run`

app/routes.py
```diff
@@ -1,2 +1,317 @@
...
+video_bp = Blueprint("video", __name__, url_prefix="/videos")
+rental_bp = Blueprint("rental", __name__, url_prefix="/rentals")
+
+@video_bp.route("/<video_id>/rentals", methods=["GET"])
+def get_rentals_by_video(video_id):
+
+    rentals = db.session.query(Rental).filter(Rental.video_id ==video_id)
+
+    if rentals is None:
+        return jsonify(None), 404
```
CheezItMan, line 22: This is fine, but it might be better to provide more feedback  to the end user.

app/routes.py
```diff
@@ -1,2 +1,317 @@
...
+
+@customer_bp.route("/<customer_id>/rentals", methods=["GET"])
+def get_rentals_by_customer(customer_id):
+
+    rentals = db.session.query(Rental).filter(Rental.customer_id ==customer_id)
+
+    if rentals is None:
+        return jsonify(None), 404
+
+    elif request.method == "GET":
```
CheezItMan, line 49: Do you need to check to see if the method is `GET`?  It's the only option.

app/routes.py
```diff
@@ -1,2 +1,317 @@
...
+
+@video_bp.route("/<video_id>/rentals", methods=["GET"])
+def get_rentals_by_video(video_id):
+
+    rentals = db.session.query(Rental).filter(Rental.video_id ==video_id)
+
+    if rentals is None:
+        return jsonify(None), 404
+
+    elif request.method == "GET":
```
CheezItMan, line 24: Do you need to check to see if the method is `GET`?  It's the only option.

app/routes.py
```diff
@@ -1,2 +1,317 @@
...
+        if request_body["customer_id"] and request_body["video_id"]:
+            
+            returned_rental = Rental(customer_id=request_body["customer_id"],
+                                video_id=request_body["video_id"])
+
+            customer = Customer.query.get(request_body["customer_id"])
+            video = Video.query.get(request_body["video_id"])
+
+            if not customer or not video:
+                response = "Not found tho"
```
CheezItMan, line 79: You might want to provide feedback on what wasn't found

app/routes.py
```diff
@@ -1,2 +1,317 @@
...
+                "release_date": video.release_date,
+                "title": video.title,
+                "due_date": rental.due_date
+            })
+
+        return jsonify(response), 200
+
+@rental_bp.route("/check-in", methods=["POST"])
+def return_video():
+    if request.method == "POST":
```
CheezItMan, line 66: ???

app/routes.py
```diff
@@ -1,2 +1,317 @@
...
+                                video_id=request_body["video_id"])
+
+            customer = Customer.query.get(request_body["customer_id"])
+            video = Video.query.get(request_body["video_id"])
+
+            if not customer or not video:
+                response = "Not found tho"
+                return jsonify(response), 404
+
+            elif customer.videos_checked_out_count != 0:
```
CheezItMan, line 82: Just in case it's somehow negative.

```suggestion
            elif customer.videos_checked_out_count > 0:
```

app/routes.py
```diff
@@ -1,2 +1,317 @@
...
+                response = {"customer_id": returned_rental.customer_id,
+                            "video_id": returned_rental.video_id,
+                            "videos_checked_out_count": customer.videos_checked_out_count,
+                            "available_inventory": video.available_inventory}
+                
+                return jsonify(response), 200
+            
+            else:
+                response = "No available inventory"
+            return jsonify(response), 400
```
CheezItMan, line 98: Because the `if` and `elif` have returns, you don't need the else here.

```suggestion
            response = "No available inventory"
            return jsonify(response), 400
```

app/routes.py
```diff
@@ -1,2 +1,317 @@
...
+        new_rental = Rental(customer_id=request_body["customer_id"],
+                            video_id=request_body["video_id"],
+                            due_date = now + timedelta(7))
+
+        customer = Customer.query.get(request_body["customer_id"])
+        video = Video.query.get(request_body["video_id"])
+
+        if not customer or not video:
+            response = "Not found tho"
+            return jsonify(response), 404
```
CheezItMan, line 124: Just note that you're doing a lot of checking to see if specific fields are valid in your request body.  Maybe having a function to check on that would be good. It might dry up your code a bit.

app/routes.py
```diff
@@ -1,2 +1,317 @@
...
+            response = "No available inventory"
+        return jsonify(response), 400
+
+    else:
+        response = "Must include name_id and video_id"
+        return jsonify(response), 400
+
+
+@customer_bp.route("", methods=["GET", "POST"], strict_slashes=False)
+def handle_customer():
```
CheezItMan, line 152: Just a note, you seem to switch a bit between route functions that handle one method and path, and like this which do multiple methods.  Sticking with one style would be more consistent.  

If you're worrying about drying up your code, you can accomplish that with decorators.

app/routes.py
```diff
@@ -1,2 +1,317 @@
...
+    
+    elif request.method == "POST":
+
+        request_body = request.get_json()
+
+        if all(key in request_body for key in ("name", "postal_code", "phone")):
+            new_customer = Customer(name=request_body["name"], 
+                                    postal_code=request_body["postal_code"],
+                                    phone=request_body["phone"],
+                                    registered_at=now)
```
CheezItMan, line 182: Just noting another point where you're making sure the body of the request is valid.  This would make a good case for a helper function.

app/routes.py
```diff
@@ -1,2 +1,317 @@
...
+
+        response = customer.json_response()
+
+        return jsonify(response)
+
+    elif request.method == "PUT":
+
+        form_data = request.get_json()
+
+        try:
```
CheezItMan, line 213: Nice experimentation with try-except blocks.

CheezItMan (commented):
Saejin, this was a pretty solid submission.  I mostly had comments regarding the organization of the code.  

I suggest putting functions handling "Business Logic" of your application in the models classes.  This would handle operations like checking out and checking in videos.  You can also create methods to do things like looking up rentals by video etc.  So the models handle business logic and database operations and `routes.py` only handles coordinating the request and response and use the model methods.  

You could also use some helper functions to validate the request and detect 404 errors, missing fields etc.

Otherwise this was a solid submission.  Well done.

PR #94, opened 2021-11-15
Spruce - Sandra & Michelle 

DynamicDev3000:


requirements.txt
```diff
@@ -1,20 +1,25 @@
...
 autopep8==1.5.5
+black==21.10b0
 blinker==1.4
 certifi==2020.12.5
 chardet==4.0.0
 click==7.1.2
 Flask==1.1.2
 Flask-Migrate==2.6.0
 Flask-SQLAlchemy==2.4.4
+gunicorn==20.1.0
```
anselrognlie, line 12: We do need to add gunicorn here for deployment, but where did the other changes come from?

migrations/versions/d9c729291cd8_rental_model.py
```diff
@@ -0,0 +1,54 @@
+"""rental model
```
anselrognlie, line 1: Make sure that your messages give a good description of the overall migration. This migration creates all three of the models in the application, not just rental.

.gitignore
```diff
@@ -67,6 +67,7 @@ db.sqlite3-journal
 # Flask stuff:
 instance/
 .webassets-cache
+.flashenv
```
anselrognlie, line 70: Nice use of an ignored .flaskenv file to set your development flags

app/__init__.py
```diff
@@ -32,5 +32,13 @@ def create_app(test_config=None):
...
 
     #Register Blueprints Here
+    from .routes import customers_bp
+    app.register_blueprint(customers_bp)
+
+    from .routes import videos_bp
+    app.register_blueprint(videos_bp)
+
+    from .routes import rentals_bp
+    app.register_blueprint(rentals_bp)
```
anselrognlie, line 42: 👍 

One thing we can do to help our routes file from getting too large is to consider making multiple files containing routes separated by the resource type. We might have a `routes` folder instead of a `routes.py` file, and inside that folder (along with a `__init__.py`) we could have a file per resource, so `customer.py`, `video.py`, and `rental.py`. Where each would have a blueprint and endpoints for that resource. When we have one blueprint per file, we often name the blueprint simply `bp` rather  than including the resource name as part of it.

Then here, we could import and register the blueprints like
```python
    from .routes import customer, video, rental
    app.register_blueprint(customer.bp)
    app.register_blueprint(video.bp)
    app.register_blueprint(rental.bp)
```

app/models/customer.py
```diff
@@ -1,4 +1,26 @@
 from app import db
+from datetime import date
+from app.models.rental import Rental
 
 class Customer(db.Model):
-    id = db.Column(db.Integer, primary_key=True)
\ No newline at end of file
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
```
anselrognlie, line 6: According to the SqlAlchemy docs about [autoincrement](https://docs.sqlalchemy.org/en/14/core/metadata.html#sqlalchemy.schema.Column.params.autoincrement):

> The default value is the string "auto" which indicates that a single-column primary key that is of an INTEGER type with no stated client-side or python-side defaults should receive auto increment semantics automatically; all other varieties of primary key columns will not.

This means that here, since the primary is not a compound key, the type is integer, and there's no specified default value, the desired autoincrement behavior will apply even without listing it. It doesn't hurt anything to list it, but it's not required.

app/models/customer.py
```diff
@@ -1,4 +1,26 @@
...
+    name = db.Column(db.String)
+    postal_code = db.Column(db.String)
+    phone = db.Column(db.String)
+    registered_at = db.Column(db.DateTime, default=date.today())
```
anselrognlie, line 10: Keep in mind that the default value for `nullable` is `True`. When adding column definitions to our models, we should consider whether it makes sense to allow a record to exist without that data, such as a Customer without a name, a Video without a title, or a Rental without a due date. It's true that we can add checks for this in our logic that adds entries, but we should try to leverage as much validation help from the database as possible. If we tell the database that certain columns are required (not nullable), then we'll encounter an error if we accidentally try to create a row without a value for those columns (which is a good thing)!

anselrognlie: Consider using `datetime.utcnow()` rather than `date.today()`. When working with dates and time, it's usually a good idea to pass them around and store them in UTC, and only convert them to local time when needed for display (like in a UI). Working with dates and times is tricky!

app/models/customer.py
```diff
@@ -1,4 +1,26 @@
...
+    #rentals = db.relationship('Rental', backref = 'customer', lazy=True)
+    videos = db.relationship("Video", secondary="rental", backref="customers")
```
anselrognlie, line 12: Notice that your code doesn't use this relationship in the logic anywhere. Using the `secondary` attribute on a relationship is more appropriate when we have a simple many-to-many relationship, such as with a pure join table. In that case, a relationship like this would allow SQLAlchemy to automatically manage the join table between the two models being linked and we could establish those links by using the relationship like a collection, by appending, removing, etc. In that case, the `backref` attribute supplies the equivalent name to attach to the model on the other end of the relationship. Since the local name here is `videos`, a closer match in semantics for the other side might be `customers` rather than `customer_table`.

However, in this application, our rental relationship between videos and customers isn't a pure join table. It has its own information that also needs to be tracked, and which SQLAlchemy can't automatically manage (such as the checked in status, and due date). In this case, using reciprocal one-to-many relationships can bring us some of the benefits of moving around the relationship links, while still leaving it to us to explicitly set up the rental model ourselves. Consider a configuration like the following:

```python
    rentals = db.relationship("Rental", backref="customer")
```

If we had a customer model in the variable `customer`, then we could access the rentals associated with the customer as `customer.rentals`. If there were a similar relationship on the video side, then to get the video title of a particular customer rental, we could write code like `customer.rentals[0].video.title`. SQLAlchemy takes care of using the foreign key information to fill in the relationships.

app/models/customer.py
```diff
@@ -1,4 +1,26 @@
...
+
+#helper function to return dict
+    def to_dict(self):
+        return {
+            "id" : self.id,
+            "name" : self.name,
+            "postal_code" : self.postal_code,
+            "phone" : self.phone,
+            "registered_at" : self.registered_at
+        }
```
anselrognlie, line 22: 👍 

Nice method to convert the model into a dict for returning. We have a variety of endpoint outputs that we need to handle in this project, but having a general-purpose dictionary converter is a great start!

app/models/customer.py
```diff
@@ -1,4 +1,26 @@
...
+
+    def videos_rental_query_by_customer(self):
+        rental_query = Rental.query.filter_by(customer_id = self.id)
+        return rental_query.count()
```
anselrognlie, line 26: Other than the dictionary helper, this is probably the single most useful helper method to add to the customer: providing a way to easily look up the rentals outstanding for this customer. But you should also add a filter to restrict to rentals that are still checked out, otherwise this will also find rentals that have been checked in already (which is helpful for getting the customer history, but not when determining their current checked out count).

Also, this isn't being called anywhere. 😭  Maybe there was trouble getting it to work? This would probably be due to leaving off the checked out check.

app/models/video.py
```diff
@@ -1,4 +1,28 @@
...
+    def to_video_object(self):
+        return {
+        "id": self.id,
+        "title": self.title,
+        "release_date": self.release_date,
+        "total_inventory": self.total_inventory
+    }
+
+    def videos_rental_query(self):
+        rental_query = Rental.query.filter_by(video_id = self.id)
```
anselrognlie, line 21: As with the equivalent method in customer, make sure to include the check for the rental still being checked out. This should only report the count of currently outstanding rentals against this video, not those which have already been checked in.

app/models/video.py
```diff
@@ -1,4 +1,28 @@
...
+        return rental_query.count()
+    
+    def check_out_available_inventory(self):
+        return self.total_inventory - self.videos_rental_query()
+
+    def check_in_available_inventory(self):
+        return self.total_inventory + self.videos_rental_query()
```
anselrognlie, line 28: Good thinking again to have a helper for the available count, but I'd expect there to be only a single method which subtracts the currently number of checked out rentals for this video from the total available copies. It shouldn't matter whether this is in a check in or check out. Though we might need to make sure that we call the method after updating the rental records so that we get the correct result before returning.

app/models/rental.py
```diff
@@ -1,4 +1,12 @@
+from sqlalchemy.orm import backref
 from app import db
+from datetime import datetime, timedelta
 
 class Rental(db.Model):
-    id = db.Column(db.Integer, primary_key=True)
\ No newline at end of file
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    video_id = db.Column(db.Integer, db.ForeignKey("video.id"), primary_key=True, nullable=False)
+    customer_id = db.Column(db.Integer, db.ForeignKey("customer.id"), primary_key=True, nullable=False)
```
anselrognlie, line 8: I generally agree with your decision here to make the foreign keys not nullable. A rental without a customer or video is of questionable value. However, when a row referenced by a foreign key is deleted, postgres attempts to resolve the lost data by setting those foreign keys to NULL, which this constraint prevents, and would cause the final tests (deleting videos and customers in a rental) to fail. In this case, the secondary relationship is what is allowing those tests to pass, as it causes SQLAlchemy to [delete the rows](https://docs.sqlalchemy.org/en/14/orm/basic_relationships.html#deleting-rows-from-the-many-to-many-table) in the secondary table that depend on a deleted foreign key. This may be what we want, but on the other hand, data is very valuable to a company. Just be sure you have considered the alternatives and are intentionally choosing an approach after weighing your options.

Also, it's more usual to include the foreign keys in a compound primary key when working with a pure join table, where the compound key also signifies that there can be only a single record in the table between the pair due to primary key uniqueness constraints. In our situation, we wouldn't want that since that would mean that a customer could only ever check out a particular video once and only once! The inclusion of the autoincrementing id key prevents this effect (since the compound key will always be unique) but it also means that having the foreign keys as part of the primary key has no effect. So we can leave the primary key designation off the foreign keys.

app/models/rental.py
```diff
@@ -1,4 +1,12 @@
+from sqlalchemy.orm import backref
```
anselrognlie, line 1: We don't need this import. For simple backrefs, the relationship named parameter is sufficient. If we _do_ need the backref symbol (used to set more complex options on a backref), we can get at it though the `db` reference. Sometimes we do need to import stuff directly from SQLAlchemy, but as much as possible, we should try to bring in SQL-related symbols through the `db` ref we already have.

app/models/rental.py
```diff
@@ -1,4 +1,12 @@
...
 class Rental(db.Model):
-    id = db.Column(db.Integer, primary_key=True)
\ No newline at end of file
+    id = db.Column(db.Integer, primary_key=True, autoincrement=True)
+    video_id = db.Column(db.Integer, db.ForeignKey("video.id"), primary_key=True, nullable=False)
+    customer_id = db.Column(db.Integer, db.ForeignKey("customer.id"), primary_key=True, nullable=False)
+    due_date = db.Column(db.DateTime)
+    checked_out = db.Column(db.Boolean, default = False)
+    #customer = db.relationship("Customer", backref='rentals', lazy=True)
+    #video = db.relationship("Video", backref='rentals', lazy=True)
```
anselrognlie, line 12: These relationships _could_ be included, but since you already have a secondary relationship, they can interfere with one another. In general, I would avoid the secondary relationship in this case, since we're not working with a pure join table for rental, but that would then require looking into cascaded operations to be able to handle the case where a video or customer that was involved in a rental (active or not) is deleted. However, helpers would still be helpful for getting things like the number of active rentals for videos and customers, since this wouldn't include a check for the checked out status (note that there are ways to specify custom joins for the relationships that _could_ take the checked out field into account).

app/routes.py
```diff
@@ -0,0 +1,357 @@
...
+from datetime import datetime, timedelta
+from app.models.video import Video
+from app.models.customer import Customer
+from app.models.rental import Rental
+
+
+videos_bp = Blueprint("videos", __name__, url_prefix=("/videos"))
+
+#POST/CREATE 
+@videos_bp.route("",methods=["POST"])
```
anselrognlie, line 12: 👍 

Nice job splitting your endpoints into separate functions. There are still shared pieces of functionality in some of the endpoints that we can think about trying to reduce duplication (especially looking up a record by its id), but on the plus side, in each function, we only need to think about the logic related to that specific endpoint.

app/routes.py
```diff
@@ -0,0 +1,357 @@
...
+    
+    elif "release_date" not in request_body:
+        return{
+            "details": "Request body must include release_date."
+        }, 400
+    
+    elif "total_inventory" not in request_body:
+        return{
+            "details": "Request body must include total_inventory."
+        }, 400
```
anselrognlie, line 28: Since the outcome of any of these conditions being true is for the function to exit, these could all be `if` rather than `elif`. More interestingly, notice how each of these conditions is more or less the same. They differ only in the key being checked for in the request, and the string being used in the error message. If we put the keys into a list and iterated over it, how could we rewrite this to avoid duplicating the logic three times (or potentially more if there were other keys we needed to check!)?

app/routes.py
```diff
@@ -0,0 +1,357 @@
...
+
+    return jsonify(new_video.to_video_object()), 201
+
+#GET
+@videos_bp.route("", methods=["GET"])
+def get_videos():
+    videos = Video.query.all()
+    videos_response = []
+    for video in videos:
+        videos_response.append(video.to_video_object())
```
anselrognlie, line 46: This matches our pattern of being a good candidate for a list comprehension. When we have
```python
    result = []
    for value in values:
        result.append(value.some_method())
```

We can write this as
```python
    result = [value.some_method() for value in values]
```

In this case, that would look like:
```python
    videos_response = [video.to_video_object() for video in videos]
```

app/routes.py
```diff
@@ -0,0 +1,357 @@
...
+    return jsonify(videos_response), 200
+
+@videos_bp.route("/<video_id>",methods=["GET"])
+def get_video(video_id):
+    if video_id.isnumeric() != True:
```
anselrognlie, line 51: The most dependable way to check whether a python string represents an integral value is to try to convert it to an int in a try block, then if it raises a ValueError, we know it wasn't an int, so here, that could look like:

```python
    try:
        int(video_id)
    except ValueError:
        return {"message": "Video id provided is not a number."}, 400 
```

The string methods like `isnumeric` and `isdigit` are more about detecting classes of characters (think Unicode) than they are for determining what numeric types we might be able to convert the string value into.

app/routes.py
```diff
@@ -0,0 +1,357 @@
...
+    return{
+        "id": video.id,
+        "title": video.title,
+        "release_date": datetime.now(),
+        "total_inventory": video.total_inventory
+    }
+#PUT
+@videos_bp.route("/<video_id>",methods=["PUT"])
+def update_video(video_id):
+    video = Video.query.get(video_id)
```
anselrognlie, line 69: Notice that we should be checking the validity of the id when looking up a model in _all_ of our endpoints. If the client passed in an id to this endpoint like "hello", this code would crash.

app/routes.py
```diff
@@ -0,0 +1,357 @@
...
+#DELETE
+@videos_bp.route("/<video_id>", methods=["DELETE"])
+def delete_video(video_id):
+
+    video = Video.query.get(video_id)
+
+    if video is None:
+        return {"message": f"Video {video_id} was not found"}, 404
+
+    db.session.delete(video)
```
anselrognlie, line 101: Just to reiterate, the deletion of a customer that is being used in a rental is also going to delete any rental record that references this customer due to the secondary relationship that was configured between video and customer using the rental table. Without that relationship, once a customer had been involved in a rental (currently checked out or not), it would become impossible to delete the customer from the database since the foreign key relationships in rental were set to be non-nullable.

app/routes.py
```diff
@@ -0,0 +1,357 @@
...
+    db.session.commit()
+
+    return jsonify({"id": video.id}), 200
+
+###Customer
```
anselrognlie, line 106: The basic CRUD operations for customers are largely the same as for videos, so look for places where the feedback on video endpoints could apply to customer endpoints as well.

app/routes.py
```diff
@@ -0,0 +1,357 @@
...
+        return jsonify(details="Request body must include customer id and video id"), 400
+
+    customer_id = rental_request["customer_id"]
+    video_id = rental_request["video_id"]
+
+    try: 
+        video_id = int(video_id)
+        customer_id = int(customer_id)
+    except:
+        return jsonify(None), 400
```
anselrognlie, line 223: 👍 

Nice check for id validity. This is common to overlook (and some other endpoints could still use a similar check). For that reason, it can be really useful to wrap checks like this into helper functions so that rather than using the raw query.get, we use our "smarter" lookup helper.

app/routes.py
```diff
@@ -0,0 +1,357 @@
...
+
+    customer = Customer.query.get(customer_id)
+    video = Video.query.get(video_id)
+    if customer is None or video is None:
+        return jsonify(None), 404
+
+    num_currently_checked_out = Rental.query.filter_by(video_id=video.id, checked_out=True).count()
+    available_inventory = video.total_inventory - num_currently_checked_out
```
anselrognlie, line 231: Consider moving the logic for calculating the checked out count and available inventory to helper methods in the Video or Rental model (or updating and using the method you already added!).

app/routes.py
```diff
@@ -0,0 +1,357 @@
...
+
+    if available_inventory == 0:
+        return jsonify(message="Could not perform checkout"), 400
+
+    new_rental_check_out= Rental(
+        customer_id=customer.id,
+        video_id=video.id,
+        due_date=(datetime.now() + timedelta(days=7)),
```
anselrognlie, line 239: Consider using `utcnow()` rather than `now()`, especially if there's no timezone information on the datetime in the database. When working with dates and time, it's usually a good idea to pass them around and store them in UTC, and only convert them to local time when needed for display (like in a UI). Working with dates and times is tricky!

app/routes.py
```diff
@@ -0,0 +1,357 @@
...
+        checked_out=True
+    )
+
+    db.session.add(new_rental_check_out)
+    db.session.commit()
+
+    available_inventory -= 1
+
+    videos_customer_checked_out = Rental.query.filter_by(customer_id=customer.id, checked_out=True).count()
```
anselrognlie, line 248: Consider moving the logic for calculating the checked out count to a helper method in the Customer or Rental model (or again, updating and making use of the method you already added!).

app/routes.py
```diff
@@ -0,0 +1,357 @@
...
+        video_id = int(video_id)
+        customer_id = int(customer_id)
+    except:
+        return jsonify(None), 400
+
+    customer = Customer.query.get(customer_id)
+    video = Video.query.get(video_id)
+
+    if customer is None or video is None:
+        return jsonify(None), 404
```
anselrognlie, line 278: I know that Task List had some odd error behavior for not found records resulting in the response body being `None`. That seems to be what this helper is modeled on. Personally, I found the Task List behavior unsatisfying, and I think it may have been a bit of a technical mix up in that project, and not something we'd really want to replicate here. None of the tests checking for responses look for an explicit `None`, so we could provide more helpful error reporting than we had in Task List.

app/routes.py
```diff
@@ -0,0 +1,357 @@
...
+    
+    rental = Rental.query.filter_by(customer_id=customer.id, video_id=video.id, checked_out=True).first()
```
anselrognlie, line 280: 👍 

Great job getting the checked out rentals between the customer and video!

Consider ordering the rentals by the due date. A customer could have multiple rentals for the same video outstanding, and it would be nice to guarantee the oldest rental is considered to be returned first.

app/routes.py
```diff
@@ -0,0 +1,357 @@
...
+        customer_id = int(customer_id)
+    except:
+        return jsonify(None), 400
+    
+    customer = Customer.query.get(customer_id)
+    
+    if customer == None:
+        return jsonify(message=f"Customer {customer_id} was not found"), 404
+
+    rental_list = Rental.query.filter_by(customer_id=customer.id, checked_out=True)
```
anselrognlie, line 313: Good job including the checked out check here. It's that requirement that makes the default relationship not do what we would like. But we could added a helper that includes the checked out value.

app/routes.py
```diff
@@ -0,0 +1,357 @@
...
+
+    list_of_dicts = []
+
+    for rental in rental_list:
+        
+        video = Video.query.get(rental.video_id)
```
anselrognlie, line 319: Even though we can't use the default relationships to get the current checked out rentals for customers or videos, they would still help us here, by letting us write `rental.video.title` or `rental.video.release_date` directly, without needing to do the lookup ourselves. As mentioned elsewhere, those relationships can interact badly with secondary relationships, which is why I would tend to avoid the secondary relationship in this project and use the regular one-to-may relationships.

anselrognlie (commented):
Well done!

All your tests are passing and your code is very readable. Though notice that you forked from AdaGold rather than Ada-C16. Be sure to take with where you clone from. The AdaGold version can generate noisy emails for instructors when changes like PRs occur. 🙏

Try looking for repetitive parts of the code and moving them to helper methods. For example, the logic around checking whether an id is compatible with the get method could reside in a custom method to retrieve model records more safely. Or the rental queries could be moved into helpers on customer or video (or a neutral 3rd part class) to avoid needing to remember all the details of the query each time it needs to be written. You definitely started down this path, and identified some great helpers, but didn't take them to completion.

We can also think about how to simplify our routes (and get closer to the single responsibility principle) by reducing the amount of logic directly written in them. The main job of a route should be to get the data from the request, then call helper methods to do the real work (business logic), and finally take the results and package them to be sent back to the client. Currently, there's business logic (such as how to update the database to reflect checking out and in) in your routes. Think about moving that to methods in your existing model classes, or even creating entirely new classes that represent sequences of actions carried out at the level of business logic. Consider adding a class like `RentalManager`, that could have `check_out` and `check_in` methods to manage those workflows.

Overall, nicely done! 🎉

PR #95, opened 2021-11-15
Pine - Roslyn & Areeba

AreebaQ:


app/routes.py
```diff
@@ -0,0 +1,444 @@
...
+            db.session.commit()
+    return result    
+
+#####---------------------------    Sorting   -------------------------------#####
+# Customers can be sorted by name, registered_at and postal_code
+# Videos can be sorted by title and release_date
+# Overdue rentals can be sorted by title, name, checkout_date and due_date
+
+
+def sort_titles(sort_by, entity):
```
jbieniosek, line 285: Great helper!

app/routes.py
```diff
@@ -0,0 +1,444 @@
...
+    customer_response = [customer.to_dict() for customer in customers]
+    return jsonify(customer_response), 200
+
+@customer_bp.route("", methods=["POST"])
+def create_customer():
+    request_body = request.get_json()
+    check = check_customer_video_data(method="POST", 
+                            request_body=request_body, 
+                            keys=customer_keys)
+    return check if check else create_customer(request_body)
```
jbieniosek, line 50: Great use of helpers!

app/routes.py
```diff
@@ -0,0 +1,444 @@
...
+    else:
+        return process_checkin(request_body)
+
+
+#----------------------------------------------------------------------------------#
+#---------------------------    Helper Functions    -------------------------------#
+#----------------------------------------------------------------------------------#
+
+#####---------------------    Model Object Creation   -------------------------#####
+def create_customer(request_body):
```
jbieniosek, line 210: I recommend renaming this, "create_customer_helper" for example.  Using the same name creates some readability issues, and can also cause weird behavior issues in Python.

app/routes.py
```diff
@@ -0,0 +1,444 @@
...
+    return sorted
+
+    # ex: entities = MyEntity.query.order_by(desc(MyEntity.time)).limit(3).all()
+
+#####------------------------    Data Checking   ----------------------------#####
+def check_customer_video_data(method=None, request_body=None, keys=None, id=None, model=None, entity=None):
+    if method == "POST" or method == "PUT":
+        for key in keys:
+            if key not in request_body.keys():
+                return make_response({"details": f"Request body must include {key}."}, 400)
```
jbieniosek, line 324: Using abort here would could simplify the code around this function:
```suggestion
                abort(make_response({"details": f"Request body must include {key}."}, 400))
```
The calling function could then just call check_customer_video_data(...) rather than capturing and testing the return value from the function.

app/routes.py
```diff
@@ -0,0 +1,444 @@
...
+    new_rental = create_rental(request_body, due_date)
+    return make_response(new_rental.to_dict(checked_out=video.inventory_checked_out, available_inventory=inventory_available), 200)
+
+def process_checkin(request_body):
+    video_id = request_body["video_id"]
+    customer_id = request_body["customer_id"]
+    video = Video.query.get(video_id)
+    video.inventory_checked_out -= 1
+    inventory_available = calculate_inventory_available(video)
+    rentals = Rental.query.filter(Rental.customer_id==customer_id).all() #
```
jbieniosek, line 268: Multiple filter params can be passed here, which would eliminate the need for the for loop below:
```suggestion
    rentals = Rental.query.filter(Rental.customer_id==customer_id, Rental.video_id==video_id).all() #
```
Note, this would still return a list, but unless a user is allowed to check out multiple copies of the same video, there would be a max of one element in the list.  To avoid the list, something like `.one()` or `.one_or_none()` could be used (https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.one_or_none).

jbieniosek (commented):
Fantastic work on this project Roslyn & Areeba!  I really like the way you used helper functions to simplify the code in your routes, as a result all of the routes are very streamlined and easy to read.  I also appreciate the way you organized the helper functions, your code was very clear and easy to walk through.  My only recommendation would be to pull some of these helper functions out into separate files and use Python file namespaces to help organize things.  This project is green!

